PR #9, opened 2019-09-26
fix: have button connected to one element

niekvanstaveren:


Caele:
I don't get it? won't it add more buttons each time it renders?

niekvanstaveren:
> I don't get it? won't it add more buttons each time it renders?

Actually yes it does. I was trying to fix the issue where you can only have one button object on a sheet

src/components/action-button.js
```diff
@@ -10,9 +10,8 @@ export default function ActionButton({ layout }) {
       color = component.fontColor;
     }
   }
-  const button = document.createElement('button');
   button.setAttribute('style', `color: ${color}; 'font-size': ${fontSize}`);
-  button.appendChild(document.createTextNode('Not React'));
+  button.innerHTML = 'Not React';
```
Caele: should avoid using innerHTML. It's not an issue here, but as we want to insert an user inputted expression, it can get nasty if not escaped correctly. There is a comment on it here: https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML

PR #17, opened 2019-10-07
fix: styling logic

haxxmaxx:


src/ext.js
```diff
@@ -20,6 +20,7 @@ export default function ext(/* env */) {
                   component: 'string',
                   ref: 'style.label',
                   translation: 'Label',
+                  placeholderTranslation: 'My button',
```
haxxmaxx: Do we want a placeholder or a default value? Either way it should be possible to set the label empty so you get a blank button. Maybe solved if we allow expressions.

niekvanstaveren: I guess a default value would be easier then. But let's check with Christofer tomorrow!

PR #18, opened 2019-10-07
Apply bookmark and Clear all selections

niekvanstaveren:


Caele:
Can you plan to separate even the property panel logic for actions into its own module? So that in the ext file you just do actionModule.propertyDef() or something.

niekvanstaveren:
I created a new user story for separating the default properties from the ext.js file. Closing this PR now since there is another one up with more recent changes including sequential handling of the actions

PR #22, opened 2019-10-10
feat: QLIK-97976, remaining selection actions

haxxmaxx:
TODO:
- ~~decide on when we should override lock~~ where engine call supports it, DONE
- ~~decide on if we should merge `selectField` and `selectValues`~~ yes, DONE
- ~~decide on if we should removed chained actions~~ yes, DONE


src/components/action-button.js
```diff
@@ -19,7 +19,7 @@ export default function ActionButton({ layout, button, Theme, engineApp, context
       const { actions } = layout;
       actions.forEach(action => {
         const findAction = actionsList.find(act => act.value === action.actionType);
-        actionList.push(findAction.promise({ engineApp, ...action }));
+        findAction && actionList.push(findAction.call({ engineApp, ...action }));
```
niekvanstaveren: Let's add check on call here as well:

typeof findAction.call === 'function'  or just findAction.call

src/ext.js
```diff
@@ -65,6 +65,14 @@ export default function ext(/* env */) {
...
               show: data => checkShow(data, 'softLock'),
             },
+            value: {
+              // TODO: expressions
+              type: 'string',
+              ref: 'value',
+              component: 'string',
+              label: 'Value',
+              show: data => checkShow(data, 'value'),
+            },
```
niekvanstaveren: Should we create a new user story for expressions? I know Tobias was looking into this as well

src/utils/actions.js
```diff
@@ -1,18 +1,25 @@
+// TODO: put in seperate file for testing. Create util fil for other functions as well?
```
niekvanstaveren: If we want to use it in different places then yes. Otherwise we could export this function for testing as well

src/utils/actions.js
```diff
@@ -1,18 +1,25 @@
...
+
 const actions = [
   {
     value: 'applyBookmark',
     label: 'Apply a bookmark',
     group: 'bookmark',
-    promise: ({ engineApp, bookmark }) => async () => {
-      await engineApp.applyBookmark(bookmark);
+    call: ({ engineApp, bookmark }) => async () => {
+      bookmark && (await engineApp.applyBookmark(bookmark));
```
niekvanstaveren: Yes like these checks!

src/utils/actions.js
```diff
@@ -68,71 +79,146 @@ const actions = [
...
-  {
-    value: 'setVariable',
-    label: 'Set variable value',
-    group: 'variables',
-  },
+  // {
+  //   value: 'setVariable',
+  //   label: 'Set variable value',
+  //   group: 'variables',
+  // },
```
niekvanstaveren: Just commenting for now or did you find a limitation?

src/components/action-button.js
haxxmaxx: sure, on it

src/utils/actions.js
haxxmaxx: Can't think of other uses yet. I'll export it and test it

src/utils/__tests__/actions.spec.js
```diff
@@ -0,0 +1,227 @@
...
+    await actionObj.getActionCall({ engineApp, bookmark })();
+    expect(engineApp.applyBookmark).to.have.been.calledWith(bookmark);
+  });
+
+  it('should NOT call applyBookmark when no bookmark', async () => {
+    const actionObj = actions.find(action => action.value === 'applyBookmark');
+    bookmark = null;
+    await actionObj.getActionCall({ engineApp, bookmark })();
+    expect(engineApp.applyBookmark).to.not.have.been.called;
+  });
```
niekvanstaveren: Awesome!

src/utils/actions.js
```diff
@@ -100,90 +100,56 @@ const actions = [
...
-        await fieldObj.selectAll();
+        await fieldObj.selectAll(softLock);
       }
     },
-    requiredInput: ['field'],
+    requiredInput: ['field', 'softLock'],
   },
   {
     value: 'selectValues',
     label: 'Select multiple values in a field',
```
niekvanstaveren: I think we should change this label to 'Select value(s) in a field', since we are removing the single one

src/utils/actions.js
niekvanstaveren: We can remove the comment now I think

haxxmaxx: yes just haven't pushed it yet

src/utils/__tests__/style-formatter.spec.js
```diff
@@ -1,26 +1,13 @@
-import styleFormatter from '../utils/style-formatter';
+import styleFormatter from '../style-formatter';
+import defaultValues from '../../components/__tests__/default-button-props';
```
haxxmaxx: @niekvanstaveren I don't like rewriting code, but this import might be questionable 🙃

PR #24, opened 2019-10-14
fix: add babel polyfill to run in nebula

niekvanstaveren:


niekvanstaveren:
@miralemd In order for the object to run in Nebula I added babel-polyfill. It was required since we use async functions. Could you have a look as well?

niekvanstaveren:
Ah cool! I will check it out :)

PR #25, opened 2019-10-14
feat: Set variable action

haxxmaxx:
Only changes the value for the session, so not setting properties.

### DIFF
--- a/src/ext.js
+++ b/src/ext.js
@@ -24,6 +24,7 @@ export default function ext(/* env */) {
           addTranslation: 'Add Item',
           items: {
             actionType: {
+              // TODO: searchable dropdown
               type: 'string',
               ref: 'actionType',
               component: 'dropdown',
@@ -31,6 +32,7 @@ export default function ext(/* env */) {
               options: actions,
             },
             bookmark: {
+              // TODO: searchable dropdown
               type: 'string',
               ref: 'bookmark',
               component: 'dropdown',
@@ -45,6 +47,7 @@ export default function ext(/* env */) {
               show: data => checkShow(data, 'bookmark'),
             },
             field: {
+              // TODO: searchable dropdown
               type: 'string',
               ref: 'field',
               component: 'dropdown',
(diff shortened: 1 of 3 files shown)


src/ext.js
```diff
@@ -58,12 +58,12 @@ export default function ext(/* env */) {
...
-            softLock: {
-              type: 'boolean',
-              ref: 'softLock',
-              label: 'Overwrite locked selections',
-              defaultValue: false,
-              show: data => checkShow(data, 'softLock'),
+            variable: {
+              type: 'string',
+              ref: 'variable',
+              component: 'string',
```
haxxmaxx: I think we can keep this as a string component, since we don't need expressions for the name.
Or maybe it should be a dropdown, since you can only change variables not create new ones.

niekvanstaveren: Not sure if you would need an expression for the name, maybe we should check that. Otherwise for now it's okay I think. If we go for a dropdown then we definitely don't have the option for expression anymore. And if we go that road then we should probably have a searchable dropdown anyway

haxxmaxx: just checked and the old button has an expression component

src/utils/actions.js
```diff
@@ -180,11 +180,22 @@ const actions = [
...
+    label: 'Set variable value',
+    group: 'variables',
+    getActionCall: ({ app, variable, value }) => async () => {
+      if (variable && value) {
+        try {
+          const variableObj = await app.getVariableByName(variable);
+          await variableObj.setStringValue(value);
+        } catch (e) {
+          // no-op
+        }
```
haxxmaxx: Otherwise we would need to fetch all the variables and make sure it exists.
Seems to be no side effects of this

niekvanstaveren: Would it be enough to check for existence of variableObj?

haxxmaxx: No. The error is coming from `getVariableByName` if an incorrect variable name is sent, which breaks the object :/

niekvanstaveren: Ah very strict! :)

PR #26, opened 2019-10-14
fix: default properties setup

niekvanstaveren:


niekvanstaveren:
@miralemd In order for this to run in Nebula I think we need babel-polyfill to be added. Can you have a look as well?

src/object-properties.js
```diff
@@ -1,8 +1,29 @@
...
+  /**
+   * @typedef {object}
+   */
+  style: {
+    /**
+     * @type {string}
+     */
+    label: '',
+  },
 };
```
miralemd: I would suggest adding `actions: []` here to make sure it exists when the object is created

src/utils/style-formatter.js
```diff
@@ -13,15 +13,12 @@ export default {
   getStyles(style, Theme) {
     // TODO: use constants for default values?
     let styles = 'width: 100%;height: 100%;font-weight: bold;';
-
-    if (style) {
+    if (Theme) {
```
haxxmaxx: Should add a test for when Theme is undefined, given that could actually happen, otherwise remove the check.

haxxmaxx: Also `Theme && (palette = Theme.getCurrent().properties.palettes.ui[0].colors)` might look nicer

Caele: wrap the theme stuff in its own file as well, there is a chance that that palettes doesn't exist on the theme properties, so we need a way to fall back to something in that case. And for the case where Theme doesn't exist at all, like when outside of sense.

src/utils/theme-resolver.js
```diff
@@ -12,5 +12,13 @@ const themeResolver = {
     }
     return 'none';
   },
+  getPalette: Theme => {
```
haxxmaxx: Nice!

PR #30, opened 2019-10-15
feat: added background image

eliseeborn:


src/ext.js
```diff
@@ -132,7 +149,89 @@ export default function ext(/* env */) {
                     color: null,
                   },
                 },
-              ],
+                useBackgroundImage: {
+                  ref: 'style.background.isUsed',
+                  type: 'boolean',
+                  translation: 'properties.backgroundImage.use',
+                  component: 'switch',
+                  defaultValue: false,
```
niekvanstaveren: @Caele Do you think we will ever have the option to set a background image from the Theme? In that case we cannot set a defaultValue here, but it feels like quite a stretch

niekvanstaveren:
@eliseeborn When I switched between fit to width and fit to height I see some errors popping up. Could you have a look at that?

eliseeborn:
@niekvanstaveren So, it's due to that align-matrix assumes that background.position is in $scope.data, and here it's in like layout.style.background.position - so when align-matrix tries to change position because the old position is no longer allowed due to changing the size, it throws an error and doesn't change the position.
I'm not 100% sure about the structure of the scope, so I don't really know how to fix this.

haxxmaxx:
Since this is open source we want to keep out internal stuff like jira numbers (Miralem told us yesterday, thats why there are older commits with jira numbers). When working in the client it's great to add them tho 👍

eliseeborn:
ok, I'll remove it when I merge then :)

haxxmaxx:
Great, change the name of the PR as well

src/ext.js
Caele: no that would be strange, the images right now has to come from within sense, you cannot link to an external one (which a theme one would almost be)

src/ext.js
```diff
@@ -5,6 +5,23 @@ export const checkShow = (data, field) => {
   return act && act.requiredInput && act.requiredInput.indexOf(field) !== -1;
 };
 
+export const getValue = (data, reference, defaultValue) => {
+  const steps = reference.split('.');
+  let dataContainer = data;
```
haxxmaxx: Just call the argument `dataContainer`?

eliseeborn: it's a util function from client originally, but it doesn't seem to make any difference

niekvanstaveren: Changing the variable names might make it easier to understand. Also I would suggest breaking this out into another file so we keep ext.js cleaner. (Yes that probably also goes for checkShow and another one that I am adding for navigation :))

src/ext.js
```diff
@@ -126,16 +142,98 @@ export default function ext(/* env */) {
...
+                  show(data) {
+                    return (
+                      getValue(data, 'style.background.isUsed')
+                      && getValue(data, 'style.background.url.qStaticContentUrlDef.qUrl')
+                      && getValue(data, 'style.background.size') !== 'fill'
+                    );
+                  },
+                  currentSize(data) {
+                    return getValue(data, 'style.background.size');
+                  },
```
niekvanstaveren: What does this one do?

src/utils/__tests__/style-formatter.spec.js
```diff
@@ -42,5 +43,55 @@ describe('style-formatter', () => {
...
+          isUsed: true,
+          url: {
+            qStaticContentUrl: {
+              qUrl: someUrl
+            }
+          }
+        }
+      };
+      const formattedStyle = styleFormatter.getStyles(style, Theme);
+      expect(formattedStyle.includes(`background-image: url('http://localhost:4848${someUrl}')`)).to.be.true;
```
niekvanstaveren: Interesting that this works for the pipeline as well

src/utils/style-formatter.js
```diff
@@ -13,10 +33,22 @@ export default {
...
     palette = themeResolver.getPalette(Theme);
     styles += formatColorProperty('color', style.fontColor, '#ffffff');
     styles += formatProperty('font-size', style.fontSize ? `${style.fontSize}px` : '12px');
     styles += formatColorProperty('background-color', style.backgroundColor, '#3F8AB3');
+    if (style && style.background && style.background.isUsed) {
+      let bgUrl = style.background.url.qStaticContentUrl.qUrl;
+      bgUrl.replace('http://localhost:4848/', '').replace(/^\.\.\//i, '/');
+      bgUrl = bgUrl.replace(/"/g, '\\"');
+      bgUrl = bgUrl.replace(/'/g, "\\'");
+      const bgImg = `http://localhost:4848/${(bgUrl[0] === '/' ? bgUrl.substr(1) : bgUrl)}`;
```
niekvanstaveren: This won't work for environments not running on localhost

src/ext.js
eliseeborn: i think this is what align-matrix uses to watch the changes in size?

src/utils/style-formatter.js
```diff
@@ -13,10 +33,21 @@ export default {
...
+
     palette = themeResolver.getPalette(Theme);
     styles += formatColorProperty('color', style.fontColor, '#ffffff');
     styles += formatProperty('font-size', style.fontSize ? `${style.fontSize}px` : '12px');
     styles += formatColorProperty('background-color', style.backgroundColor, '#3F8AB3');
+    if (style && style.background && style.background.isUsed) {
+      let bgUrl = style.background.url.qStaticContentUrl.qUrl;
+      bgUrl.replace(/^\.\.\//i, '/');
+      bgUrl = bgUrl.replace(/"/g, '\\"');
+      bgUrl = bgUrl.replace(/'/g, "\\'");
```
niekvanstaveren: @Caele I think these checks were copied from the text and image implementation. What we don't have here is the addition of the base path url. Would this be a requirement? Also more in general how do we make sure we are good on a security perspective?

PR #31, opened 2019-10-15
feat: Dropdown for variables + expression for values

haxxmaxx:
Adding a checkbox to show/hide system variables in the list

Also removed `enigmaModel` since we don't want to expose that.

### DIFF
--- a/src/__tests__/ext.spec.js
+++ b/src/__tests__/ext.spec.js
@@ -1,3 +1,4 @@
+/* eslint-disable prefer-destructuring */
 import ext, { checkShow } from '../ext';
 
 describe('ext', () => {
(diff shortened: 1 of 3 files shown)


src/__tests__/ext.spec.js
```diff
@@ -22,7 +23,86 @@ describe('ext', () => {
   });
 
   describe('ext', () => {
-    const result = ext();
-    expect(result).to.be.an('object');
+    const props = ext();
```
haxxmaxx: Maybe a bit overkill, but I think `option` is important to test at least.

src/ext.js
```diff
@@ -65,16 +65,32 @@ export default function ext(/* env */) {
...
+                  }));
+              },
               show: data => checkShow(data, 'variable'),
             },
+            showSystemVariables: {
+              type: 'boolean',
+              ref: 'showSystemVariables',
+              label: 'showSystemVariables',
+              defaultValue: false,
+              show: data => checkShow(data, 'showSystemVariables'),
```
niekvanstaveren: Not sure if we need a specific one for showSystemVariables. We always show this when variable is selected right?

haxxmaxx: ah true, i fix

PR #32, opened 2019-10-17
feat: add navigation for action button

niekvanstaveren:


niekvanstaveren:
@eliseeborn and @haxxmaxx Please start having a look. Addition of navigation and I moved some code around as well. For now still a draft since I want to add some unit test still.

niekvanstaveren:
@Caele Basic navigation functionality. However it assumes that firstsheet and lastsheet will be present. I think we now said that they won't right?

src/utils/navigation-actions.js
```diff
@@ -0,0 +1,125 @@
...
+    requiredInput: ['story'],
+  },
+  {
+    label: 'Open a website / eMail',
+    value: 'openWebsite',
+    navigationCall: async ({ websiteUrl, sameWindow }) => {
+      try {
+        if (websiteUrl) {
+          const url =
+            websiteUrl.startsWith('https://') || websiteUrl.startsWith('http://') || websiteUrl.startsWith('mailto://')
```
miralemd: `startsWith` is a fairly new method on `String` - will not work in IE, might wanna polyfill this

miralemd: or use a regex: `websiteUrl.match(/^(https?|mailto):\/\//)`

Caele:
> @Caele Basic navigation functionality. However it assumes that firstsheet and lastsheet will be present. I think we now said that they won't right?

no, assume that those will not be part of the API

src/utils/navigation-actions.js
niekvanstaveren: Ah thanks! Will change to regex

src/utils/navigation-actions.js
```diff
@@ -0,0 +1,125 @@
...
+  },
+  {
+    label: 'Go to first sheet',
+    value: 'firstSheet',
+    navigationCall: async ({ senseNavigation }) => {
+      try {
+        await senseNavigation.firstSheet();
+      } catch (error) {
+        // no-op
+      }
```
Caele: why use a try/catch?

PR #35, opened 2019-10-22
feat: Expressions for label, font size

haxxmaxx:


src/utils/__tests__/style-formatter.spec.js
```diff
@@ -44,16 +44,22 @@ describe('style-formatter', () => {
       expect(formattedStyle.includes(`font-size: ${someSize}px`)).to.be.true;
     });
 
+    it('should return default font size for incorrect value', () => {
```
haxxmaxx: This does not add to the coverage but I think an important case to test.

src/utils/style-formatter.js
```diff
@@ -45,7 +46,10 @@ export default {
       bgUrl = bgUrl.replace(/'/g, "\\'");
       styles += formatProperty('background-image', `url('${bgUrl}')`);
       styles += formatProperty('background-size', backgroundSize[style.background.size] || backgroundSize.auto);
-      styles += formatProperty('background-position', backgroundPosition[style.background.position] || backgroundPosition.topLeft);
+      styles += formatProperty(
```
haxxmaxx: This was auto formatted. But do we need the default value? Since there are default values for `background.position` and `background.size`

eliseeborn: it could probably be removed so that it's just specified in one place, I don't know which is best

niekvanstaveren: No we should not have it in two places. I suggest removing the default values here.

haxxmaxx: Yea I agree. Keeping it aligned with the other styling defaults.

PR #36, opened 2019-10-23
feat: adding translation to action button

niekvanstaveren:


niekvanstaveren:
We can only merge this if we also merge changes in sense-client. https://github.com/qlik-trial/sense-client/pull/9456

PR #37, opened 2019-10-24
feat: Alternate states

haxxmaxx:
Maybe should update the unit test and mock `qStateName`, check if `getField()` is called with it. Not sure if we want sense specific things in there if we don't have to (tests run fine now without the mock).

### DIFF
--- a/src/utils/__tests__/actions.spec.js
+++ b/src/utils/__tests__/actions.spec.js
@@ -45,10 +45,6 @@ describe('actions', () => {
       };
     });
 
-    // afterEach(() => {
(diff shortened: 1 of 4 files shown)


src/utils/__tests__/actions.spec.js
```diff
@@ -43,6 +45,10 @@ describe('actions', () => {
       };
     });
 
+    // afterEach(() => {
+
+    // });
+
```
niekvanstaveren: Please remove. Otherwise looks good!

PR #40, opened 2019-10-24
test: update rendering test with more styles

aoj2:
Rendering test for action button with label, bgColor and bgImage.


aoj2:
Are we planning to change default borders soon? Then I might wait for that.... before merge

niekvanstaveren:
> Are we planning to change default borders soon? Then I might wait for that.... before merge

Yes Elise is actually working on that. Not sure when we will have it available in a new release though. I think we should be able to release the button more often from now on

PR #41, opened 2019-10-25
fix: moved enable condition properties out of style

eliseeborn:


src/utils/style-formatter.js
```diff
@@ -42,18 +40,18 @@ export default {
...
+    disabled && (styles += formatProperty('opacity', 0.4));
 
     if (style.background && style.background.isUsed) {
       let bgUrl = style.background.url.qStaticContentUrl.qUrl;
       bgUrl.replace(/^\.\.\//i, '/');
       bgUrl = bgUrl.replace(/"/g, '\\"');
       bgUrl = bgUrl.replace(/'/g, "\\'");
       styles += formatProperty('background-image', `url('${bgUrl}')`);
-      styles += formatProperty('background-size', backgroundSize[style.background.size] || backgroundSize.auto);
+      styles += formatProperty('background-size', backgroundSize[style.background.size]);
```
haxxmaxx: I've removed the default values, so we need to keep this or put the default values back.

eliseeborn: ah, I got confused then - I'll put them back. :)

src/utils/__tests__/style-formatter.spec.js
```diff
@@ -96,25 +97,19 @@ describe('style-formatter', () => {
...
     });
 
     it('should have set opacity for disabled button', () => {
-      style = {
-        useEnabledCondition: true,
-        enabledCondition: 0
-      };
-      const formattedStyle = styleFormatter.getStyles(style, Theme);
+      disabled = true;
+      const formattedStyle = styleFormatter.getStyles(style, disabled, Theme);
```
niekvanstaveren: You could set true inside the function here as well. Then you don't have to reset disabled in the next test

eliseeborn: like this?
`const formattedStyle = styleFormatter.getStyles(style, true, Theme);`

niekvanstaveren: exactly! In that way you don't change the default values and start every unit test in the same way. Another way would be to have a beforeEach function, but that seems a bit much for this single test

PR #42, opened 2019-10-25
feat: adding conversion to action button

niekvanstaveren:
Thanks to Christian who pointed out that we can handle this in the action button as well :)

TODO:
- Convert expression for field selection

### DIFF
--- a/src/utils/conversion.js
+++ b/src/utils/conversion.js
@@ -1,4 +1,4 @@
-function convertAction(action, newProperties) {
+export function convertAction(action, newProperties) {
   const newAction = {
     actionType: action.actionType,
     bookmark: action.selectedBookmark,
@@ -22,7 +22,7 @@ function convertAction(action, newProperties) {
       newAction.actionType = 'selectValues';
       break;
     case 'selectAndLockField':
-      newProperties.actions.push({ ...newAction, cId: 'test', actionType: 'selectValues' });
+      newProperties.actions.push({ ...newAction, cId: `${newAction.cId}1`, actionType: 'selectValues' });
       newAction.actionType = 'lockField';
       break;
     default:
@@ -31,7 +31,7 @@ function convertAction(action, newProperties) {
   return newProperties.actions.push(newAction);
 }
 
-function convertNavigation(oldType) {
+export function convertNavigation(oldType) {
   switch (oldType) {
     case 'gotoSheet':
       return 'goToSheet';
--- a/src/utils/conversion.js
+++ b/src/utils/conversion.js
@@ -1,0 +1,100 @@
+function convertAction(action, newProperties) {
+  const newAction = {
+    actionType: action.actionType,
+    bookmark: action.selectedBookmark,
+    // Will need to convert field selected by expression
+    field: action.selectedField,
+    // Need to convert expression for variable
+    variable: action.variable,
+    value: action.value,
+    softLock: action.softLock,
+    cId: action.cId,
+  };
+  switch (action.actionType) {
+    case 'clearOther':
+      newAction.actionType = 'clearAllButThis';
+      break;
+    case 'unlockAllAndClearAll':
+      newAction.actionType = 'clearAll';
+      newAction.softLock = true;
+      break;
+    case 'selectField':
+      newAction.actionType = 'selectValues';
+      break;
+    case 'selectAndLockField':
+      newProperties.actions.push({ ...newAction, cId: 'test', actionType: 'selectValues' });
+      newAction.actionType = 'lockField';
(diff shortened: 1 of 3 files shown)


src/utils/conversion.js
```diff
@@ -0,0 +1,100 @@
...
+      break;
+    case 'unlockAllAndClearAll':
+      newAction.actionType = 'clearAll';
+      newAction.softLock = true;
+      break;
+    case 'selectField':
+      newAction.actionType = 'selectValues';
+      break;
+    case 'selectAndLockField':
+      newProperties.actions.push({ ...newAction, cId: 'test', actionType: 'selectValues' });
```
niekvanstaveren: Almost forgot that I added a hard coded test here. Good that I'm writing some unit tests :)

src/utils/conversion.js
```diff
@@ -0,0 +1,99 @@
...
+      break;
+    case 'unlockAllAndClearAll':
+      newAction.actionType = 'clearAll';
+      newAction.softLock = true;
+      break;
+    case 'selectField':
+      newAction.actionType = 'selectValues';
+      break;
+    case 'selectAndLockField':
+      newProperties.actions.push({ ...newAction, cId: `${newAction.cId}1`, actionType: 'selectValues' });
```
niekvanstaveren: @haxxmaxx and @eliseeborn what do you think of the cId here? We make two actions from the old select and lock field action. However every action has a required cId. We could of course also try to create a new Id here, but then we would like to do that in the same way as has been done for the other Ids

haxxmaxx: Hmm, I think they are randomly generated, but not unique. So we don't need to keep track of all the old ones to create new. But generating a cId would require putting that logic in the button repo I guess

niekvanstaveren: Yes and that is something I would like to avoid. However I say that cId is required, but have not tried saving without the cId. Let me try that actually

haxxmaxx: OK, even if it's possible I think we should run it by Tobias

niekvanstaveren: So setting the cId to null seems to work. This ID is used to have the order in place for the items so I looked at what happens when you change the order. In that case a new Id is generated for the ones that don't have one yet. So my suggestion would be to set to null. @Caele what do you think?

haxxmaxx: So the ordering of the items is based on random generated cIds? We probably need to do something about the order anyway

Caele: Did we still need the cId?

niekvanstaveren: We don't need it, but it is automatically generated when you use an array in the property panel

Caele: ok, then it should not matter

src/utils/conversion.js
```diff
@@ -0,0 +1,99 @@
...
+            story: props.selectedStory,
+            websiteUrl: props.websiteUrl,
+            sameWindow: props.sameWindow,
+          };
+          break;
+        case 'qStateName':
+          newProperties[key] = exportedFmt.properties[key];
+          break;
+        default:
+          newProperties.qLayoutExclude.disabled[key] = exportedFmt.properties[key];
```
haxxmaxx: Stuff that is not used in the new button?

src/utils/conversion.js
```diff
@@ -0,0 +1,99 @@
+export function convertAction(action, newProperties) {
+  const newAction = {
+    actionType: action.actionType,
+    bookmark: action.selectedBookmark,
+    // Will need to convert field selected by expression
```
haxxmaxx: Are you planning to do this in this PR?

src/utils/conversion.js
niekvanstaveren: Yes, I'm not really sure if this is needed and what it is used for. But the conversions in sense seem to do the same. I think it had something to do with reverting back, but I'm not really sure

src/utils/conversion.js
niekvanstaveren: Not really. This depends on how the new component will handle things as well I guess

haxxmaxx: Oh ok, then it can wait yes

src/utils/conversion.js
haxxmaxx: I'd say ask someone in V&I, or @Caele for that matter

Caele: yea, it is for when converting back, but in this case you cannot do that so do not transfer over things that are not used.

niekvanstaveren: I'll remove it then

PR #44, opened 2019-10-28
fix: have pointer when hover button

niekvanstaveren:


haxxmaxx:
set `border: none` while you're at it

PR #46, opened 2019-10-30
feat: Expression for color

haxxmaxx:
Used the logic from `Color` in sense. Slimmed it down to what we need. A bunch of details that I am unsure of, see comments


src/ext.js
```diff
@@ -232,12 +232,46 @@ export default function ext({ translator }) {
...
+                          translation: 'Off',
+                        },
+                      ],
+                    },
+                    fontColorExpression: {
+                      component: 'string',
+                      type: 'string',
+                      ref: 'style.fontColorExpression',
+                      translation: 'Expression',
+                      expression: 'optional',
```
haxxmaxx: This makes the expression evaluate so that `hsl(x,x,x) -> RGB(y,y,y)`which we need. But this also makes it possible to set incorrect strings, without a `=`.

src/utils/color-resolver.js
```diff
@@ -1,5 +1,33 @@
...
+    // argb
+    matches = /^argb\(\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*,\s*(\d{1,3})\s*\)$/i.exec(input);
+    if (matches) {
+      const a = Math.round((matches[1] / 255) * 100) / 100;
+      return `rgba(${matches[2]},${matches[3]},${matches[4]},${a})`;
+    }
+    // hex
+    matches = /^#([A-f0-9]{2})([A-f0-9]{2})([A-f0-9]{2})$/i.exec(input);
+    if (matches) {
+      return input;
```
haxxmaxx: Do we want to align the returns? the `rgba()` format would work for all.

src/utils/color-utils.js
```diff
@@ -0,0 +1,62 @@
...
+    matches = /^#([A-f0-9]{2})([A-f0-9]{2})([A-f0-9]{2})$/i.exec(input);
+    if (matches) {
+      return input;
+    }
+    // css color
+    const color = input && CSSColors[input.toLowerCase()];
+    if (color) {
+      const a = color.a !== undefined ? color.a : 1;
+      return `rgba(${color.r},${color.g},${color.b},${a})`;
+    }
```
niekvanstaveren: @Caele Max is away for some days, so maybe you can help here. Is there any reason why we need to convert the color to rgba? This is done the same way as in the client, but I would say it would work just putting the color directly in the style as well. 
Could it be that this is for security reasons, to make sure it is an actual color that is set? Still no need to convert to rgba I would say though

haxxmaxx: If we don't this look up the color, there is no way of setting a default color. We return 'none' if there are none of the cases are true. But it could just be a look-up table kind of, returning true if it is a correct color, then not converting it to rgb(a).

haxxmaxx: No, but you can type whetever you want in the expression field. If we don't have the list of colors, we have no way of checking if the color name is correct

haxxmaxx:
@niekvanstaveren Have you checked other colors that might be missing?

niekvanstaveren:
> @niekvanstaveren Have you checked other colors that might be missing?

Yes css colors usually stay the same. However, this color was added a couple years ago. I wanted to check the difference because I saw that we had transparent listed which is not really a css color. Then I stumbled on this: http://lists.w3.org/Archives/Public/www-style/2014Jun/0312.html

src/utils/color-utils.js
Caele: don't know exactly why we check against a list of colors, could be that we want to be able to set transparency for example. Or being extra careful security wise. Either way, this looks fine I think.

PR #47, opened 2019-10-30
feat: Text styling and align

haxxmaxx:
Screenshot 2019-11-06 at 16 22 43

Modified the component so that we can select multiple values: https://github.com/qlik-trial/sense-client/pull/9564


src/ext.js
```diff
@@ -238,6 +238,59 @@ export default function ext(/* env */) {
...
+                      icon: 'align_center',
+                      value: 'center',
+                    },
+                    {
+                      component: 'icon-item',
+                      icon: 'align_right',
+                      value: 'right',
+                    },
+                  ],
+                  defaultValue: 'center',
```
eliseeborn: The default value is both here and in style-formatter, maybe one should be removed?

haxxmaxx: Good catch. Yes, we need to keep it here though, since we need to show the correct value in the pp. Actually that should go for background defaults as well. Just noticed that the position does not have a value until you choose one.

src/ext.js
```diff
@@ -272,6 +272,71 @@ export default function ext({ translator }) {
...
+                  ref: 'style.textStyle',
+                  translation: 'Text Style',
+                  horizontal: true,
+                  multipleSelect: true,
+                  items: [
+                    {
+                      component: 'icon-item',
+                      icon: 'bold',
+                      value: 'bold',
+                      translation: 'Bold',
```
haxxmaxx: Strings needed

niekvanstaveren: Probably best to put in the translation strings before merging. Even if we don't add them to the client yet it would be easier to spot not translated strings

src/ext.js
```diff
@@ -272,6 +272,71 @@ export default function ext({ translator }) {
...
+                      component: 'icon-item',
+                      icon: 'underline',
+                      value: 'underline',
+                      translation: 'Underline',
+                      labelPlacement: 'bottom',
+                    },
+                  ],
+                  defaultValue: {
+                    bold: true,
+                  },
```
niekvanstaveren: How is this stored? I assume every value has its own boolean? We should have these defaults in object.properties

src/ext.js
niekvanstaveren: Should all be stored in object.properties. I made a new user story for it to move all initials values there

src/utils/__tests__/style-formatter.spec.js
```diff
@@ -134,5 +136,32 @@ describe('style-formatter', () => {
...
+      };
+      const formattedStyle = styleFormatter.getStyles(style, disabled, Theme);
+      expect(formattedStyle.includes('font-style: italic')).to.be.true;
+    });
+    it('should return text-decoration: underline when underline is selected', () => {
+      style = {
+        textStyle: {
+          underline: true,
+        },
+      };
```
niekvanstaveren: In the PR for borders I changed the setup a bit for the unit test. style is reset in a beforeEach. Depending on which one we merge first there are some changes needed here. Setting the specific object on style instead of setting the whole style object

src/utils/style-formatter.js
```diff
@@ -32,8 +32,7 @@ const formatColorProperty = (path, inputColor, defaultColor, isExpression) => {
 
 export default {
   getStyles(style, disabled, Theme) {
-    // TODO: use constants for default values?
-    let styles = 'width: 100%;height: 100%;font-weight: bold;border:none;';
+    let styles = 'width: 100%;height: 100%;border: none;padding: 4px;';
```
niekvanstaveren: Why do we need the padding here?

src/ext.js
haxxmaxx: Yes, when one of them are toggled, they are added. Could add italic and underline as false as well

src/ext.js
haxxmaxx: sure I fix

src/utils/style-formatter.js
haxxmaxx: Screenshot 2019-11-06 at 15 11 17

Otherwise the text will be all the way out to the edge, UX suggested 4 px.

src/ext.js
```diff
@@ -272,6 +272,73 @@ export default function ext({ translator }) {
                     },
                   },
                 },
+                fontStyling: {
+                  component: 'item-selection-list',
+                  type: 'string',
+                  ref: 'style.textStyle',
+                  translation: 'Text Style',
```
niekvanstaveren: One last translation string

PR #53, opened 2019-11-07
fix: refactor + object props

haxxmaxx:
- Structure of `ext` , so that we get dividers in pp where we want
- Grouping stuff in the `style` object:
  - style
    - font
    - background
    - border
- Putting defaults in `object-properties`
- Simplifying `style-formatter`
- Using `object-properties` for mocking


src/utils/style-formatter.js
```diff
@@ -25,57 +24,50 @@ const backgroundPosition = {
 
 const formatProperty = (path, setting) => `${path}: ${setting};`;
 
-const getColor = (inputColor, defaultColor, isExpression) => {
-  const color = isExpression ? colorUtils.resolveExpression(inputColor) : colorUtils.resolveColor(inputColor, palette);
-  return `${color === 'none' ? defaultColor : color}`;
+const getColor = ({ useColorExpression, colorExpression, color }, defaultColor) => {
```
haxxmaxx: deconstructing ❤️

src/utils/style-formatter.js
```diff
@@ -25,57 +24,50 @@ const backgroundPosition = {
...
-      'background-color',
-      getColor(backgroundColor, primaryColor, style.useBackgroundColorExpression)
-    );
+    // enable
     disabled && (styles += formatProperty('opacity', 0.4));
     !disabled && (styles += formatProperty('cursor', 'pointer'));
-
-    if (style.background && style.background.isUsed) {
+    // font
+    styles += formatProperty('font-size', !isNaN(style.font.size) ? `${style.font.size}px` : '12px');
```
haxxmaxx: @miralemd All default values in style (e.g. `style.font.size`) are defined in `object-properties`. Can I expect them to be there outside sense? Otherwise I need to put a lot of sanity checks in here.

miralemd: If they're in object-properties then they will be there when the object is created regardless of environment.

The thing you need to watch out for are properties/values that are injected by the property-panel definition/component, those will only exist in Sense.

haxxmaxx: Awesome. Making sure the unit tests are using `object-properties` for mocking.

src/utils/__tests__/style-formatter.spec.js
```diff
@@ -1,24 +1,21 @@
...
+    const { Theme, layout } = defaultProperties;
     const disabled = false;
     let element;
 
     beforeEach(() => {
-      style = {
-        border: { isUsed: false },
-        textAlign: 'center',
-      };
+      style = JSON.parse(JSON.stringify(layout.style));
```
niekvanstaveren: I checked if this was necessary and it is, just don't understand really why

src/utils/__tests__/style-formatter.spec.js
```diff
@@ -29,61 +26,96 @@ describe('style-formatter', () => {
...
+
+    it('should return specified background color from expression', () => {
+      style.background.colorExpression = someColorExpression;
+      style.background.useColorExpression = true;
+      const formattedStyle = styleFormatter.getStyles({ style, disabled, Theme });
+      expect(formattedStyle.includes(`background-color: ${someColorExpression}`)).to.be.true;
+    });
+
+    it('should return default background color when color is none', () => {
+      style.background.color = { index: 0 };
```
niekvanstaveren: Don't need this declaration of color here.

src/object-properties.js
```diff
@@ -37,26 +37,108 @@ const properties = {
...
+        },
+      },
+      /**
+       * @type {string}
+       */
+      size: 'auto',
+      /**
+       * @type {string}
+       */
+      position: 'topLeft',
```
niekvanstaveren: I think a center position as default for the image would be better actually. Maybe discuss with Christofer

src/object-properties.js
```diff
@@ -37,26 +37,108 @@ const properties = {
...
      */
     label: 'Button',
+    /**
+     * @type {object}
+     */
+    font: {
+      /**
+       * @type {string}
+       */
+      size: '',
```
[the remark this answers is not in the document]
haxxmaxx: No. We are using a string component, which enables expressions

src/object-properties.js
haxxmaxx: sure, I'll change

src/utils/__tests__/style-formatter.spec.js
haxxmaxx: setting an object to a already existing object, makes it a reference, not a copy. So the `beforeEach` does not reset `style`, since `layout.style` also changes

src/utils/style-formatter.js
```diff
@@ -1,7 +1,6 @@
 import colorUtils from './color-utils';
 
 let palette;
```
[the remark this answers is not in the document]
haxxmaxx: Sure, I guess that is better than declaring it globaly

PR #54, opened 2019-11-08
feat: responsive font size

niekvanstaveren:
First draft at responsive font size. Let me know what you think :)

EDIT: Ready for review!


src/utils/style-formatter.js
```diff
@@ -79,4 +78,18 @@ export default {
 
     return styles;
   },
+  setFontSize(button, style) {
```
haxxmaxx: Could destructureto label and fontSize

src/components/action-button.js
```diff
@@ -12,14 +12,13 @@ export const runActions = async actionList => {
 export default function ActionButton({ layout, button, Theme, app, context, senseNavigation, element }) {
   const { style, qStateName } = layout;
   const disabled = layout.useEnabledCondition && layout.enabledCondition === 0;
-  const formattedStyles = styleFormatter.getStyles({ style, disabled, Theme, element });
+  const formattedStyles = styleFormatter.getStyles({ style, disabled, Theme, element, button });
```
haxxmaxx: button not used?

haxxmaxx:
Will be confilcts with the refactoring changes. Mostly becasuse `style.fontSize` -> `style.font.size` etc

niekvanstaveren:
> Will be confilcts with the refactoring changes. Mostly because `style.fontSize` -> `style.font.size` etc

That is okay. I will wait with merging this until refactoring is done. This was more prototype to see if the solution is the way we want to go. Unit tests would need to be rewritten for this one as well since I have added another html element here

src/utils/style-formatter.js
niekvanstaveren: Yes everything needs to be cleaned up still

src/components/action-button.js
niekvanstaveren: No, noticed that button was not used in the getStyles function. Maybe the setting of the label was in there before but I don't know

src/components/__tests__/root.spec.js
```diff
@@ -1,5 +1,8 @@
 import defaultValues from '../../__tests__/default-button-props';
-import { render } from '../root';
+import { render, teardown } from '../root';
+
+// Empty function should be test when filled
+teardown();
```
niekvanstaveren: Probably taking it too far here, but would be great if we can have all files except index.js fully covered

src/ext.js
```diff
@@ -242,11 +242,13 @@ export default function ext({ translator }) {
...
                   items: {
                     fontSize: {
-                      component: 'string',
-                      type: 'string',
+                      component: 'slider',
+                      type: 'number',
                       ref: 'style.font.size',
                       translation: 'properties.fontSize',
-                      expression: 'optional',
+                      min: 0,
```
niekvanstaveren: Maybe minimum should be bigger than 0, but not really sure

src/object-properties.js
```diff
@@ -42,9 +42,9 @@ const properties = {
      */
     font: {
       /**
-       * @type {string}
+       * @type {number}
        */
-      size: '',
+      size: 50,
```
niekvanstaveren: Not sure about the default, but 50 (%) looks a bit better than 100 I think

src/components/__tests__/root.spec.js
haxxmaxx: So you are just running the function to get the coverage? I actually think we should wait to write an actual test

niekvanstaveren: I don't think we will implement the teardown function. I'm actually thinking of removing this completely

niekvanstaveren: @haxxmaxx So removed the teardown function. Pretty sure we are not going to implement anything here

src/ext.js
haxxmaxx: 1 works fine. Don't think we need to be able to set it to 0, since you can just remove the label if you want to have no label.

src/object-properties.js
haxxmaxx: Looks good to me to, has UX any thoughts?

niekvanstaveren: Not sure, but I will show it to Christofer tomorrow and ask what he thinks.

niekvanstaveren: Christofer is also good on 50%!

src/components/__tests__/root.spec.js
haxxmaxx: Great!

src/utils/__tests__/style-formatter.spec.js
```diff
@@ -121,6 +108,16 @@ describe('style-formatter', () => {
       expect(formattedStyle.includes('background-repeat: no-repeat')).to.be.true;
     });
 
+    it('should return no settings shen url is missing', () => {
```
haxxmaxx: spelling

src/components/root.js
```diff
@@ -1,8 +1,8 @@
 import ActionButton from './action-button';
+import styleFormatter from '../utils/style-formatter';
 
-export function teardown() {}
-
-export function render(element, props) {
+export default function render(element, props) {
   const button = ActionButton({ element, ...props });
   element.appendChild(button);
+  styleFormatter.setFontSizeAndFamily({ button, ...props });
```
haxxmaxx: Why do we have to call it here and not inside ActionButton?

src/utils/style-formatter.js
```diff
@@ -75,4 +73,22 @@ export default {
 
     return styles;
   },
+  setFontSizeAndFamily({ button, Theme, layout }) {
+    const { style } = layout;
+    function setFontsize(textElement, newFontsize) {
```
haxxmaxx: I'd say place this outside like the other help functions

src/components/action-button.js
```diff
@@ -12,14 +12,13 @@ export const runActions = async actionList => {
...
   button.setAttribute('style', formattedStyles);
```
haxxmaxx: I think we can move this into `getStyles` and make it a `setStyles`, more aligned with setFont

src/components/root.js
niekvanstaveren: So the issue is that we need the button to be appended to the element so we can do our calculations on the width and height of the button (containing borders). If we don't have the button appended the size will be 0 and all calculations will be incorrect. This will be okay once the button is rendered but not on first render when you open up the page (or do a refresh)

haxxmaxx: Ok 👍

src/components/action-button.js
haxxmaxx: So remove the button argument?

src/utils/style-formatter.js
```diff
@@ -75,4 +73,22 @@ export default {
...
+    }
+    const text = button.firstElementChild;
+    text.textContent = style.label;
+    setFontsize(text, button.clientHeight);
+    let newFontsize = (button.clientHeight / text.offsetHeight) * button.clientHeight;
+    setFontsize(text, newFontsize);
+    if (text.offsetWidth + 8 > button.clientWidth) {
+      newFontsize *= (button.clientWidth - 8) / text.offsetWidth;
+    }
+    setFontsize(text, (newFontsize * style.font.size) / 100);
```
haxxmaxx: Set the font size three times?

niekvanstaveren: The calculation is done in three steps:
1. Set the font size to the height of the parent container
2. Adjust the font size based on the height ratio of the parent container and the text 
3. Adjust the font size if the text width is bigger than the parent width

niekvanstaveren: You need to actually set the font size in order to do the next calculation otherwise the size of the text box is not adjusted

haxxmaxx: fair enough :)

haxxmaxx:
FIx the comments and we are good to go :)

src/components/action-button.js
niekvanstaveren: Yes good idea. I will open a new PR for that. Will need to rewrite a lot of unit tests for that since we now check the return value in getStyles

PR #56, opened 2019-11-12
test: adding unit test for index.js and removing unused functions

niekvanstaveren:


src/index.js
```diff
@@ -29,11 +29,6 @@ export default function supernova(env) {
           senseNavigation,
         });
       },
-      resize() {},
-      willUnmount() {
-        teardown(this.element);
-      },
-      destroy() {},
     },
```
niekvanstaveren: @miralemd are these functions mandatory? I tried in sense and in nebula but seems to work just the same without these

miralemd: you don't need the methods if they're not doing anything

PR #57, opened 2019-11-13
refactor: change getstyles to setstyles

niekvanstaveren:
Leftover from the responsive text

### DIFF
--- a/src/components/action-button.js
+++ b/src/components/action-button.js
@@ -12,8 +12,7 @@ export const runActions = async actionList => {
 export default function ActionButton({ layout, button, Theme, app, context, senseNavigation, element }) {
   const { style, qStateName } = layout;
   const disabled = layout.useEnabledCondition && layout.enabledCondition === 0;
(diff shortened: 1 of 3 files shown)


haxxmaxx:
Hmm. I'm 50/50 on this (I know it was my suggestion 🙃) I thought it was `element` that we set the style to, not `button`. So this requires passing `button` and a lot more rows in the tests... Thoughts?

niekvanstaveren:
> Hmm. I'm 50/50 on this (I know it was my suggestion 🙃) I thought it was `element` that we set the style to, not `button`. So this requires passing `button` and a lot more rows in the tests... Thoughts?

Yeah it seemed a good idea and then I started changing all the unit tests. The tests are definitely nicer in the old way. @eliseeborn any thoughts?

eliseeborn:
Yeah, the code itself seems about the same either way, but the tests feel more complicated this way, so it's probably better not to change.

PR #59, opened 2019-11-18
feat: click animation

haxxmaxx:


src/components/action-button.js
```diff
@@ -12,9 +12,10 @@ export const runActions = async actionList => {
 export default function ActionButton({ layout, button, Theme, app, context, senseNavigation, element }) {
   const { style, qStateName } = layout;
   const disabled = layout.useEnabledCondition && layout.enabledCondition === 0;
+  const inEditMode = context.permissions.indexOf('interact') === -1;
```
haxxmaxx: @niekvanstaveren Does the interact permission always exist outside of edit mode or should this variable be called something else?

niekvanstaveren: That is a really good question. But changing the variable name into inInteractMode (or something like that) might be better then?

src/components/__tests__/action-button.spec.js
```diff
@@ -50,6 +50,57 @@ describe('action button', () => {
       await aButton.onclick();
       expect(button.removeAttribute).to.have.been.calledWith('disabled');
     });
+    it('should call scale and resetScale on mousedown/up', async () => {
+      const aButton = actionButton(defaults);
+      aButton.offsetHeight = 100;
+      aButton.offsetWidth = 200;
+      aButton.style = { transform: '' };
```
niekvanstaveren: Is this needed?

src/utils/style-formatter.js
```diff
@@ -35,7 +35,7 @@ const setFontsize = (textElement, newFontsize, fontFamily) => {
 
 export default {
   getStyles({ style, disabled, Theme, element }) {
-    let styles = 'width: 100%;height: 100%;padding: 4px;';
+    let styles = 'width: 100%;height: 100%;padding: 4px;transition: transform .1s ease-in-out;';
```
niekvanstaveren: Did we check with UX what speed it should have?

haxxmaxx: I showed Christofer a video of it and seemed fine with it. He probalby wants to play around with it to make sure

src/components/__tests__/action-button.spec.js
haxxmaxx: So the style object is not added unless it is not rendered for real. Wanted to mock it the way it would look in production. In this certain test it would work with only an empty style object.

src/components/action-button.js
haxxmaxx: I'll change it

PR #60, opened 2019-11-18
feat: using expression dropdown

niekvanstaveren:


src/ext.js
```diff
@@ -37,19 +37,21 @@ export default function ext({ translator }) {
               addTranslation: 'Object.ActionButton.AddAction',
               items: {
                 actionType: {
-                  // TODO: searchable dropdown
                   type: 'string',
                   ref: 'actionType',
-                  component: 'dropdown',
+                  component: 'expression-with-dropdown',
+                  translation: 'Object.ActionButton.Actions',
                   defaultValue: null,
```
haxxmaxx: This doen't give you any console errors? I had trouble with that before

niekvanstaveren: You mean the defaultValue of null? I don't think so, but we should remove it probably since it is not used at all (and therefore not giving console errors I guess)

PR #61, opened 2019-11-19
refactor: element creation in mounted function

niekvanstaveren:


haxxmaxx:
So this should fix the problem which would occur if we made a proper release to the client?

niekvanstaveren:
> So this should fix the problem which would occur if we made a proper release to the client?

At the moment there is no problem, but this is what Miralem said about that:

> it’s works today in Sense due to the way we load the extension, but with better caching that might change

haxxmaxx:
> > So this should fix the problem which would occur if we made a proper release to the client?
> 
> At the moment there is no problem, but this is what Miralem said about that:
> 
> > it’s works today in Sense due to the way we load the extension, but with better caching that might change

And this PR addresses that?

niekvanstaveren:
> > > So this should fix the problem which would occur if we made a proper release to the client?
> > 
> > 
> > At the moment there is no problem, but this is what Miralem said about that:
> > > it’s works today in Sense due to the way we load the extension, but with better caching that might change
> 
> And this PR addresses that?

Yes, @miralemd could you explain how this would change things for future caching changes?

miralemd:
There is a problem today already if you were to render more than one button in a mashup.

The exported function is only run once to initiate the supernova type. Things inside `component` is code that runs for every instance, so anything that is instance-based should be inside `component`

```js
export default function supernova() {
  /*
   * This function is only loaded and executed once
   * Any code placed here is to be considered global
   */
  return {
    qae: {},
    component: {}
  };
}
```

PR #62, opened 2019-11-20
feat: add icon feature

niekvanstaveren:


src/ext.js
```diff
@@ -536,6 +530,50 @@ export default function ext({ translator }) {
                 },
               },
             },
+            icon: {
+              type: 'items',
+              grouped: true,
```
haxxmaxx: I'm not sure this does anything. If it doesn't we can remove all uses of it in ext

src/utils/style-formatter.js
```diff
@@ -29,8 +29,13 @@ const getColor = ({ useColorExpression, colorExpression, color }, defaultColor,
   return resolvedColor === 'none' ? defaultColor : resolvedColor;
 };
 
-const setFontsize = (textElement, newFontsize, fontFamily) => {
-  textElement.setAttribute('style', `white-space: nowrap; font-size: ${newFontsize}px; font-family: ${fontFamily};`);
+const setFontsize = (textElement, newFontsize, fontFamily, addFlex, alignment) => {
```
haxxmaxx: Could only add `aligment` as parameter, then check if it exists instead of using `addFlex`

niekvanstaveren: That would work probably. However flex is added to have the icon and text vertically aligned in a nice way

haxxmaxx: Is there any reason not to use flex when not using an icon?

src/utils/style-formatter.js
```diff
@@ -79,15 +84,29 @@ export default {
   },
   setFontSizeAndFamily({ button, Theme, layout }) {
     const { style } = layout;
-    const text = button.firstElementChild;
-    text.textContent = style.label;
+    button.innerHTML = '';
+    const text = document.createElement('text');
+    const textContent = document.createElement('span');
```
haxxmaxx: I'd call this `textSpan`, aligned with `iconSpan`

src/utils/style-formatter.js
```diff
@@ -79,15 +84,29 @@ export default {
   },
   setFontSizeAndFamily({ button, Theme, layout }) {
```
haxxmaxx: As you mentioned, this function is a bit messy. Off the top of my head, can't think of a smarter way to do it. So I guess we can improve the readability at least. This function needs to be renamed for what it is actually doing, `createLabelAndIcon` or something.

src/utils/style-formatter.js
```diff
@@ -29,8 +29,13 @@ const getColor = ({ useColorExpression, colorExpression, color }, defaultColor,
...
 };
 
-const setFontsize = (textElement, newFontsize, fontFamily) => {
-  textElement.setAttribute('style', `white-space: nowrap; font-size: ${newFontsize}px; font-family: ${fontFamily};`);
+const setFontsize = (textElement, newFontsize, fontFamily, addFlex, alignment) => {
+  textElement.setAttribute(
+    'style',
+    `${
+      addFlex ? `display: flex; align-items: center; justify-content: ${alignment};` : ''
+    } font-size: ${newFontsize}px; font-family: ${fontFamily};`
```
haxxmaxx: Setting `font-family` every time the `setFontSize` is called doesn't really make sense. Maybe set it last in `setFontSizeAndFamily`. Use `textElement.style.fontSize()` to not override other styling.

src/utils/style-formatter.js
niekvanstaveren: Not really I think, but I removed the alignment from the button so we would need it here now

src/utils/style-formatter.js
```diff
@@ -77,17 +79,47 @@ export default {
...
     const { style } = layout;
-    const text = button.firstElementChild;
-    text.textContent = style.label;
-    const fontFamily = colorUtils.getFontFamily(Theme);
-    setFontsize(text, button.clientHeight, fontFamily);
+    button.innerHTML = '';
+    const text = document.createElement('text');
+    text.style.whiteSpace = 'nowrap';
+
+    // textSpan containing the text. Only set text decoration here otherwise button will get underline as well
```
haxxmaxx: @niekvanstaveren you mean the icon will be underlined? And we don't want that?

niekvanstaveren: Ah yes that is what I meant. I don't think we want that, because it looks very ugly (try it)

src/ext.js
```diff
@@ -333,7 +345,7 @@ export default function ext({ translator }) {
                         {
                           component: 'icon-item',
                           icon: 'align_left',
-                          value: 'left',
+                          value: 'flex-start',
```
niekvanstaveren: I know I changed it into flex-start and flex-end, but not so sure if we want it like that. If we would like to change the css in the future we would need to do logic based on these values instead of left and right

src/utils/__tests__/style-formatter.spec.js
```diff
@@ -194,50 +196,80 @@ describe('style-formatter', () => {
...
+    // insertBefore not working...
+    // it('should place icon first then label inside text element', () => {
+    //   style.icon.useIcon = true;
+    //   style.icon.iconType = 'someIcon';
+    //   styleFormatter.createLabelAndIcon({ Theme, button, style });
+    //   const text = button.children[0];
+    //   expect(text.children[0].style.textDecoration).to.equal('none');
+    //   expect(text.children[1].style.textContent).to.equal('Button');
+    // });
+
```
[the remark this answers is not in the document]
haxxmaxx: Don't think it is possible. When a text element is created in the test environment, insertBefore does not exist as a method. Either go back to only using appendChild or skip this. I think I'd prefer cleaner code and missing a test case, the coverage is insanely good anyways.

src/ext.js
haxxmaxx: hmm true. I can change back

src/utils/__tests__/style-formatter.spec.js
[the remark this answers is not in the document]
niekvanstaveren: I agree with the cleaner code, however we can add this function on just like we do for appendChild using unshift :)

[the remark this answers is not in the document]
niekvanstaveren: I pushed it now

src/utils/style-formatter.js
```diff
@@ -78,17 +72,41 @@ export default {
...
+    const textSpan = document.createElement('span');
+    textSpan.textContent = style.label;
+    style.font.style.underline && (textSpan.style.textDecoration = 'underline');
+    text.appendChild(textSpan);
+    // icon
+    if (style.icon.useIcon && style.icon.iconType !== '') {
+      const iconSpan = document.createElement('span');
+      iconSpan.style.textDecoration = 'none';
+      iconSpan.style.fontSize = 'inherit';
+      iconSpan.setAttribute('class', `lui-icon lui-icon--${style.icon.iconType}`);
```
miralemd: Using `lui-icon` and its types creates a very tight coupling between this component and LUI. The css and lui icon font would therefore be a requirement for this to work outside Sense, e.g. mashups. In addition, since the class names are very generic, different version of LUI/action-button could easily clash if changes have been made in either the lui css or its icon font.

This should ideally be configurable so that other font-icons could be used (material, font-awesome), at the very least it needs to be documented as a requirement.

src/utils/__tests__/style-formatter.spec.js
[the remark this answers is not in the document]
haxxmaxx: Oh didn't see that there where stuff on `document` mocked already. Thanks!

src/utils/style-formatter.js
haxxmaxx: Hmm ok. Yeah I just noticed that the integration test does not render the icon, I'm assuming that lui is missing there... Yeah I saw that there were actually alternative icons for the old button. I'd say only using lui inside sense, and either disabling icons or using other font-icons outside sense. I'll talk to the other developers.

.eslintrc.json
```diff
@@ -19,7 +19,8 @@
     "no-unused-expressions": 0,
     "object-curly-newline": 0,
     "operator-linebreak": 0,
-    "import/no-extraneous-dependencies": [2, { "devDependencies": true }]
+    "import/no-extraneous-dependencies": [2, { "devDependencies": true }],
+    "no-nested-ternary": 0
```
haxxmaxx: I think nested terneries make sense

src/components/action-button.js
```diff
@@ -10,17 +10,20 @@ export const runActions = async actionList => {
 };
 
 export default function renderButton({ layout, Theme, app, context, senseNavigation, element }) {
+  const isSense = !!senseNavigation;
```
haxxmaxx: Feels like we are checking if we are in sense somewhere else? Anyhow, the navigation api always exists inside sense right?

src/utils/style-formatter.js
haxxmaxx: Added a check so we only add the icon when in sense. We'll look at making it possible outside sense later.

src/components/action-button.js
niekvanstaveren: I think we do, but also based on navigation I think. So this is fine

PR #63, opened 2019-11-21
fix: made the default label a translation string

eliseeborn:


eliseeborn:
This only works in Sense at the moment, so I'm going to fix so that the default label in nebula isn't 'Object.ActionButton' :)

PR #67, opened 2019-11-22
feat: always fill for background image

eliseeborn:


haxxmaxx:
> Yes looks good. Will need to change this dropdown to the new dropdown, but we can do that in a different task.

Do we? I think when we have only a small fixed number of options we can use the old one.

PR #71, opened 2019-11-25
fix: default border color

eliseeborn:
waiting for confirmation from visual

### DIFF
--- a/src/utils/color-utils.js
+++ b/src/utils/color-utils.js
@@ -60,10 +60,11 @@ const colorUtils = {
     const current = getCurrentTheme(Theme);
     return (current && current.properties.fontFamily) || 'inherit';
   },
-  lightenOrDarkenColor(color, percent) {
+  getFadedColor(color) {
+    const percent = -0.1;
     const f = parseInt(color.slice(1), 16);
-    const t = percent < 0 ? 0 : 255;
-    const p = percent < 0 ? percent * -1 : percent;
+    const t = 0; // percent < 0 ? 0 : 255
+    const p = percent * -1; // percent < 0 ? percent * -1 : percent
     const R = f >> 16;
     const G = (f >> 8) & 0x00FF;
     const B = f & 0x0000FF;
--- a/src/utils/color-utils.js
+++ b/src/utils/color-utils.js
@@ -61,23 +61,21 @@ const colorUtils = {
     return (current && current.properties.fontFamily) || 'inherit';
   },
   getFadedColor(color) {
-    const percent = -0.15;
+    const percent = 0.15;
     let f; let R; let B; let G;
-    const t = 0; // percent < 0 ? 0 : 255
-    const p = percent * -1; // percent < 0 ? percent * -1 : percent
     if (color.length > 7) {
       f = color.split(',');
       const rgba = f[0].indexOf('a') !== -1;
       R = rgba ? parseInt(f[0].slice(5), 10) : parseInt(f[0].slice(4), 10);
       G = parseInt(f[1], 10);
       B = parseInt(f[2], 10);
-      return `${(rgba ? 'rgba(' : 'rgb(') + (Math.round((t - R) * p) + R)},${Math.round((t - G) * p) + G},${Math.round((t - B) * p) + B}${rgba ? `,${f[3]}` : ')'}`;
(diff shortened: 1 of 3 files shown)


eliseeborn:
just noticed there is some unexpected behavior for some colors as well, maybe a more robust method of calculating the border color is needed

src/utils/style-formatter.js
```diff
@@ -67,9 +68,10 @@ export default {
       const { width, radius } = style.border;
       const { offsetHeight, offsetWidth } = element;
       const lengthShortSide = offsetHeight < offsetWidth ? offsetHeight : offsetWidth;
+      const defaultBorderColor = colorUtils.lightenOrDarkenColor(backgroundColor, -0.15);
```
haxxmaxx: Since we are not going to use this anywhere else (for now), can be a constant inside the `lightenOrDarkenColor`.

src/utils/color-utils.js
```diff
@@ -60,6 +60,15 @@ const colorUtils = {
     const current = getCurrentTheme(Theme);
     return (current && current.properties.fontFamily) || 'inherit';
   },
+  lightenOrDarkenColor(color, percent) {
```
haxxmaxx: Use a more descriptive name, `getFadedColor` or something. At least something with get, since it returns a color.

niekvanstaveren:
Still a draft? Are we waiting for UX decision?

eliseeborn:
I changed the method so the colors changed a bit, and I need to add support for rbg and rbga colors.

src/utils/color-utils.js
```diff
@@ -60,6 +60,25 @@ const colorUtils = {
     const current = getCurrentTheme(Theme);
     return (current && current.properties.fontFamily) || 'inherit';
   },
+  getFadedColor(color) {
+    const percent = -0.15;
+    let f; let R; let B; let G;
```
niekvanstaveren: I guess you could also write `let f, R, B, G;`

src/utils/color-utils.js
```diff
@@ -60,6 +60,25 @@ const colorUtils = {
...
+    const t = 0; // percent < 0 ? 0 : 255
+    const p = percent * -1; // percent < 0 ? percent * -1 : percent
```
niekvanstaveren: You want to keep these comments in case we want to move back to percentage? Also why not set percent to a positive value and use that one?

src/utils/color-utils.js
```diff
@@ -60,6 +60,25 @@ const colorUtils = {
...
+    if (color.length > 7) {
+      f = color.split(',');
+      const rgba = f[0].indexOf('a') !== -1;
```
niekvanstaveren: What happens when I have set my color to a color starting with 'a'? (aliceblue, antiquewhite, aquamarine)

src/utils/color-utils.js
```diff
@@ -60,6 +60,25 @@ const colorUtils = {
...
+      R = rgba ? parseInt(f[0].slice(5), 10) : parseInt(f[0].slice(4), 10);
+      G = parseInt(f[1], 10);
+      B = parseInt(f[2], 10);
+      return `${(rgba ? 'rgba(' : 'rgb(') + (Math.round((t - R) * p) + R)},${Math.round((t - G) * p) + G},${Math.round((t - B) * p) + B}${rgba ? `,${f[3]}` : ')'}`;
+    }
+    f = parseInt(color.slice(1), 16);
+    R = f >> 16;
+    G = (f >> 8) & 0x00FF;
+    B = f & 0x0000FF;
+    return `#${(0x1000000 + (Math.round((t - R) * p) + R) * 0x10000 + (Math.round((t - G) * p) + G) * 0x100 + (Math.round((t - B) * p) + B)).toString(16).slice(1)}`;
```
niekvanstaveren: Do I want to know what is going on here? :)

src/utils/color-utils.js
eliseeborn: I think the linter wanted it like this, I can check

niekvanstaveren: Crazy linters

src/utils/color-utils.js
eliseeborn: That was my thought, but it might be unnecessary.

src/utils/color-utils.js
eliseeborn: bitshift magic ¯\_(ツ)_/¯

PR #72, opened 2019-11-26
fix: make sure actions get titles when converting

haxxmaxx:
The bug that Andreas found

### DIFF
--- a/src/__tests__/ext.spec.js
+++ b/src/__tests__/ext.spec.js
@@ -30,8 +30,8 @@ describe('ext', () => {
       expect(itemTitle).to.equal('actionLabel');
     });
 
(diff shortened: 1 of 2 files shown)


src/__tests__/ext.spec.js
```diff
@@ -19,7 +19,7 @@ describe('ext', () => {
     const { itemTitleRef } = props.definition.items.actions.items.actions;
 
     it('Should return action label from dropdown', () => {
-      data = { actionType: 'applyBookmark', actionLabel: '' };
+      data = { actionType: 'applyBookmark' };
```
haxxmaxx: Making sure we run one test where `actionLabel` is undefined, which it is when converting

PR #75, opened 2019-11-27
fix: convert titles

eliseeborn:
fixed so that titles and showTitles setting is the same when converting from old button

### DIFF
--- a/src/utils/__tests__/conversion.spec.js
+++ b/src/utils/__tests__/conversion.spec.js
@@ -191,5 +191,17 @@ describe('conversion', () => {
       const result = importProperties(undefined, initialProperties);
       expect(result).to.include.all.keys('qChildren', 'qProperty');
     });
+
+    it('should convert titles', () => {
+      exportedFmt.properties.showTitles = true;
+      exportedFmt.properties.title = 'hello';
+      exportedFmt.properties.subtitle = 'hello';
+      exportedFmt.properties.footnote = 'hello';
+      const result = importProperties(exportedFmt, initialProperties);
+      expect(result.qProperty.showTitles).to.be.true;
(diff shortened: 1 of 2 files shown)


src/utils/__tests__/conversion.spec.js
```diff
@@ -191,5 +191,17 @@ describe('conversion', () => {
...
+    it('should convert titles', () => {
+      exportedFmt.properties.showTitles = true;
+      exportedFmt.properties.title = 'hello';
+      exportedFmt.properties.subtitle = 'hello';
+      exportedFmt.properties.footnote = 'hello';
+      const result = importProperties(exportedFmt, initialProperties);
+      expect(result.qProperty.showTitles).to.be.true;
+      expect(result.qProperty.title).to.equal('hello');
+      expect(result.qProperty.subtitle).to.equal('hello');
+      expect(result.qProperty.footnote).to.equal('hello');
```
niekvanstaveren: Maybe use different titles so we are sure that we set them also on the correct place. hello1 hello2 hello3 would do I guess

src/utils/conversion.js
```diff
@@ -84,6 +84,18 @@ const importProperties = (exportedFmt, initialProperties) => {
...
+          break;
+        case 'title':
+          newProperties[key] = exportedFmt.properties[key];
+          break;
+        case 'subtitle':
+          newProperties[key] = exportedFmt.properties[key];
+          break;
+        case 'footnote':
+          newProperties[key] = exportedFmt.properties[key];
+          break;
```
niekvanstaveren: When using the same code you only need to declare the case statements. Like this:

```
case 'a':
 case 'b':
    do something;
    break;
```

PR #76, opened 2019-11-27
fix: property panel tweaks

eliseeborn:


haxxmaxx:
Is this the only place where we want to remove it?

niekvanstaveren:
> Is this the only place where we want to remove it?

I think we should actually go through the property panel together with Christofer and see if we are using the correct elements everywhere. Maybe also let Patrik have a look if it makes sense

eliseeborn:
> > Is this the only place where we want to remove it?
> 
> I think we should actually go through the property panel together with Christofer and see if we are using the correct elements everywhere. Maybe also let Patrik have a look if it makes sense

I'll wait with merging then

PR #77, opened 2019-11-28
fix: adding icon for dropdown

niekvanstaveren:


eliseeborn:
Any reason we can't just use the value for the icon in the dropdown as well?

niekvanstaveren:
> Any reason we can't just use the value for the icon in the dropdown as well?

In other uses of this dropdown we don't want to use value as icon. We could have used a setting in the ext file I guess, but this makes it more clear I would say.

PR #83, opened 2019-11-28
fix: made minimum font size 8px

eliseeborn:


src/utils/__tests__/style-formatter.spec.js
```diff
@@ -232,15 +232,15 @@ describe('style-formatter', () => {
       expect(text.style.justifyContent).to.equal('center');
     });
 
-    it('should set fontSize when text offsetWidth is bigger than button', () => {
```
eliseeborn: I don't fully understand what this test tested before, do we need another one to fix this?

haxxmaxx: It is the third step of the font size calculations, we check if the width of the text is too wide to fit. Then we scale it according to the width instead.

src/utils/style-formatter.js
```diff
@@ -106,7 +106,7 @@ export default {
       newFontsize *= (button.clientWidth - 8) / text.offsetWidth;
     }
     // 4. Setting final font size by scaling with the font size from the layout + other font styling
-    text.style.fontSize = `${(newFontsize * style.font.size) / 100}px`;
+    text.style.fontSize = (newFontsize * style.font.size) / 100 >= 8 ? `${(newFontsize * style.font.size) / 100}px` : '8px';
```
haxxmaxx:
```suggestion
    text.style.fontSize =`${Math.max((newFontsize * style.font.size) / 100, 8)}px`;
```

src/ext.js
```diff
@@ -272,7 +272,7 @@ export default function ext({ translator }) {
                       type: 'number',
                       ref: 'style.font.size',
                       translation: 'properties.fontSize',
-                      min: 1,
+                      min: 20,
                       steps: 1,
                       max: 100,
```
haxxmaxx: I think it would be nice to bake in the factor 100 here, so the range would be 0.2 -> 1

PR #84, opened 2019-11-29
fix: padding and alignment issues

haxxmaxx:


src/utils/__tests__/style-formatter.spec.js
```diff
@@ -251,7 +251,7 @@ describe('style-formatter', () => {
         button.children.push(child);
       };
       styleFormatter.createLabelAndIcon({ Theme, button, style });
-      expect(button.children[0].style.fontSize).to.equal('9.2px');
+      expect(button.children[0].style.fontSize).to.equal('9.399999999999999px');
```
haxxmaxx: This is unfortunate... Without the 0.94 faktor I get 10px, but apparently 10*0.95 = 0.933333... Played around with the width, did not get a more even value.

PR #89, opened 2019-12-03
fix: revert to old angular dropdowns

niekvanstaveren:


src/ext.js
```diff
@@ -88,7 +88,7 @@ export default function ext({ translator }) {
                 field: {
                   type: 'string',
                   ref: 'field',
-                  component: 'expression-with-dropdown',
+                  component: 'dropdown',
                   translation: 'Common.Field',
                   defaultValue: '',
                   dropdownOnly: true,
```
haxxmaxx: Maybe comment this?

niekvanstaveren: Sure can do that

PR #98, opened 2019-12-04
feat: example + updated readme

haxxmaxx:
~~I can't get the actions to run, something to do with async. @miralemd any why?~~

EDIT: Works now.


miralemd:
🤷‍♂️

haxxmaxx:
> maybe add a picture of the button or what the example is supposed to look like, but I don't think it's a necessary addition.

I was thinking of doing a gif of the example. But then I need to get the actions working

PR #107, opened 2020-01-14
fix: convert date

haxxmaxx:
https://jira.qlikdev.com/browse/QB-842

### DIFF
--- a/src/utils/__tests__/actions.spec.js
+++ b/src/utils/__tests__/actions.spec.js
@@ -3,6 +3,7 @@ import actions, { getValueList, checkShowAction } from '../actions';
 describe('actions', () => {
   let app;
   let fieldObject;
+  let fieldInfoObject;
   let variableObject;
   let field;
   let bookmark;
@@ -30,6 +31,9 @@ describe('actions', () => {
         selectPossible: sinon.spy(),
         toggleSelect: sinon.spy(),
       };
+      fieldInfoObject = {
+        qTags: [],
+      };
       variableObject = {
         setStringValue: sinon.spy(),
       };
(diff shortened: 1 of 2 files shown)


src/utils/actions.js
```diff
@@ -1,8 +1,13 @@
...
-  return values
-    .split(';')
-    .map(value => (isNaN(value) ? { qText: value } : { qIsNumeric: true, qNumber: Number(value) }));
-}
+export const getValueList = async (app, values, isDate) => {
+  const valuesArray = values.split(';');
+  if (isDate) {
+    for (let i = 0; i < valuesArray.length; i++) {
+      // eslint-disable-next-line no-await-in-loop
+      valuesArray[i] = await app.evaluate(`Num('${valuesArray[i]}')`);
```
Caele: this is potentially very slow, but you could try and construct a single call instead?

src/utils/__tests__/actions.spec.js
```diff
@@ -39,10 +43,12 @@ describe('actions', () => {
         back: sinon.spy(),
         forward: sinon.spy(),
         getField: sinon.stub().resolves(fieldObject),
+        getFieldDescription: sinon.stub().resolves(fieldInfoObject),
         getVariableByName: sinon.stub().resolves(variableObject),
         lockAll: sinon.spy(),
         unlockAll: sinon.spy(),
         getBookmarkList: () => [{ qData: { title: 'findMyBookmark' }, qInfo: { qId: 'myBookmarkId' } }],
+        evaluate: () => '43850;43881',
```
haxxmaxx: This is not a very good mock. The only alternative I see is to find out how engine does the conversion, seems like overkill to me...

PR #111, opened 2020-01-22
fix: undo results in empty label

niekvanstaveren:
For some reason undo does not result back in the default label set at creation of the button object but to null value. Setting empty string as default value will make sure that undo will set to empty string instead of a null.

### DIFF
--- a/src/ext.js
+++ b/src/ext.js
@@ -255,6 +255,7 @@ export default function ext({ translator }) {
                 label: {
                   component: 'string',
                   ref: 'style.label',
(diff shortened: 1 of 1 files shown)


haxxmaxx:
Can this affect backwards compability somehow? pretty easy to check if you have buttons created before this fix.

niekvanstaveren:
> Can this affect backwards compability somehow? pretty easy to check if you have buttons created before this fix.

I don't really see how, but I will check!

PR #113, opened 2020-01-31
refactor: use supernova hooks

miralemd:
- use new hooks APIs from `@nebula.js/supernova`
- always use the provided theme


src/__tests__/default-button-props.js
```diff
@@ -1,14 +1,10 @@
...
-        palettes: { ui: [{ colors: ['none', 'color1', 'color2'] }] },
-        dataColors: { primaryColor: 'myPrimaryColor' },
-        fontFamily: 'myFont',
-      },
-    }),
+const defaultValues = (sandbox) => ({
+  theme: {
+    getDataColorSpecials: () => ({ primary: 'myPrimaryColor' }),
+    getColorPickerColor: sandbox ? sandbox.stub() : () => '',
+    getStyle: sandbox ? sandbox.stub() : () => '',
```
haxxmaxx: Nice!

src/utils/style-formatter.js
```diff
@@ -23,27 +23,26 @@ const backgroundPosition = {
...
 const formatProperty = (path, setting) => `${path}: ${setting};`;
 
-const getColor = ({ useColorExpression, colorExpression, color }, defaultColor, palette) => {
+const getColor = ({ useColorExpression, colorExpression, color }, defaultColor, theme) => {
   const resolvedColor = useColorExpression
     ? colorUtils.resolveExpression(colorExpression)
-    : colorUtils.resolveColor(color, palette);
-  return resolvedColor === 'none' ? defaultColor : resolvedColor;
+    : theme.getColorPickerColor(color);
+  return !resolvedColor || resolvedColor === 'none' ? defaultColor : resolvedColor;
```
haxxmaxx: Is this where you were having trouble with none?

miralemd: kind of - the default palette used in Sense has `"none"` as the first 'color' which causes all color indexes to be shifted by one - that's fine though since we should be able to handle palettes of different sizes anyway

haxxmaxx:
LGTM. Would be nice to have the `css-colors` and `resolveColors` in nebula as well, since we are using them for the org chart as well.

miralemd:
@haxxmaxx I'm planning on it - we already use `d3-color` for the [contrasting helper](https://github.com/qlik-oss/nebula.js/blob/master/apis/theme/src/contraster/luminance.js#L1), just need to figure out how to expose an API for it

PR #123, opened 2020-02-27
fix: make overflow visible to stop trimming of italics

eliseeborn:
Screenshot 2020-02-27 at 10 20 22
making the overflow visible might be risky, but it doesn't seem to cut into the borders since we added the padding


haxxmaxx:
Try writing a really long text so the font size becomes the min value, the text will not show the ellipsis just go all the way to the edge

eliseeborn:
> Try writing a really long text so the font size becomes the min value, the text will not show the ellipsis just go all the way to the edge

ah true - thought it was too easy

eliseeborn:
Screenshot 2020-02-27 at 11 10 21
yeah, I guess this is less than ideal

src/utils/style-formatter.js
```diff
@@ -110,12 +110,21 @@ export default {
       newFontsize *= button.clientWidth / text.offsetWidth;
     }
     // 4. Setting final font size by scaling with the font size from the layout + other font styling
-    if (hasIcon) {
+    if (hasIcon && style.font.style.italic) {
+      text.style.fontSize = `${Math.max(newFontsize * style.font.size * 0.84, 8)}px`;
+      text.children[0].style.marginRight = `${text.offsetWidth * 0.04}px`;
+      text.children[1].style.marginRight = `${text.offsetWidth * 0.04}px`;
```
haxxmaxx: Still need this when overflow is visible?

eliseeborn: it makes large italic text look more centered, but to be fair if the text is very small then the offset is too much anyway
it's probably better without it I think

PR #138, opened 2020-04-16
feat: add selectMatchingValues action

haxxmaxx:
Adding this action would solve a bug and make sure we have full feature parity.

I also want to revert the fix I made for dates, since it only covers some cases. Not sure if that is possible since it was delievered in MS36. Will look into how this should be released.

### DIFF
--- a/src/utils/actions.js
+++ b/src/utils/actions.js
@@ -102,7 +102,7 @@ const actions = [
   },
   {
     value: 'selectMatchingValues',
-    translation: '$Select matching values in a field',
+    translation: 'Object.ActionButton.SelectMatchingValues',
     group: 'selection',
     getActionCall: ({ app, qStateName, field, value, softLock }) => async () => {
       if (field && value) {
--- a/src/utils/actions.js
+++ b/src/utils/actions.js
@@ -101,6 +101,19 @@ const actions = [
     requiredInput: ['field', 'value', 'softLock'],
   },
   {
+    value: 'selectMatchingValues',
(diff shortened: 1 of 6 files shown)


src/utils/actions.js
```diff
@@ -100,6 +100,18 @@ const actions = [
     },
     requiredInput: ['field', 'value', 'softLock'],
   },
+  {
+    value: 'selectMatchingValues',
+    translation: '$Select matching values in a field',
```
niekvanstaveren, line 105: Needs proper translation string

haxxmaxx: Yup. started the discussion in the PS channel to come up with a good name.

haxxmaxx:
@Caele I can't see any problems with backwards compatibility, since we are not changing anything to existing actions. Everything else looks good from an architectural pov?

Caele:
> @Caele I can't see any problems with backwards compatibility, since we are not changing anything to existing actions. Everything else looks good from an architectural pov?

What happens if you use the new action and then downgrade?

haxxmaxx:
> > @Caele I can't see any problems with backwards compatibility, since we are not changing anything to existing actions. Everything else looks good from an architectural pov?
> 
> What happens if you use the new action and then downgrade?

Good question, I'll try it

haxxmaxx:
@Caele the action shows up, but missing a lot of info, and does, and does nothing when clicking. What do we want to happen? Not showing it at all might lead to some severe misunderstandings when the button just does nothing without any feedback in older versions.

PR #148, opened 2020-05-04
chore: version 1.5.0

niekvanstaveren:


haxxmaxx:
for stardust?

niekvanstaveren:
> for stardust?

Yes this is for stardust. Creating a minor version here because we would need patches for earlier versions of Sense

PR #163, opened 2020-07-15
fix: mailto link without slashes

eliseeborn:


src/utils/navigation-actions.js
```diff
@@ -83,7 +83,7 @@ const navigationActions = [
     navigationCall: async ({ websiteUrl, sameWindow }) => {
       try {
         if (websiteUrl) {
-          const url = websiteUrl.match(/^(https?|mailto):\/\//) ? websiteUrl : `http://${websiteUrl}`;
+          const url = websiteUrl.match(/^(https?:\/\/|mailto:(\/\/)?)/) ? websiteUrl : `http://${websiteUrl}`;
```
niekvanstaveren, line 86:
```suggestion
          const url = websiteUrl.match(/^(https?:\/\/|mailto:)/) ? websiteUrl : `http://${websiteUrl}`;
```

I guess this should be enough as well

eliseeborn: I'm adding a unit test for mailto as well, so I'll fix that at the same time :)

niekvanstaveren: Perfect!

PR #183, opened 2020-12-04
feat: add doReload action

Caele:
Adds doReload as an Action, with the partial flag possible.
My prettier settings did some things, but the lint passed so..

### DIFF
--- a/src/utils/actions.js
+++ b/src/utils/actions.js
@@ -2,13 +2,13 @@ export const getValueList = async (app, values, isDate) => {
   let valuesArray = values.split(';');
   if (isDate) {
     let dateExpression = '';
-    valuesArray.forEach(date => {
+    valuesArray.forEach((date) => {
       dateExpression += `Num('${date}')&';'&`;
     });
     const convertedDates = await app.evaluate(dateExpression.slice(0, -5));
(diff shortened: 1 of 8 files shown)


niekvanstaveren (approved):
Looks good. I think lint has been updated quite a bit since we did this work 😄

src/utils/actions.js
```diff
@@ -218,10 +218,22 @@ const actions = [
     },
     requiredInput: ['variable', 'value'],
   },
+  {
+    value: 'doReload',
+    translation: 'Object.ActionButton.DoReload',
+    group: 'reload',
+    getActionCall: ({ app, partial }) => async () => {
+      const e = await app.doReload(0, !!partial, false);
+      if (e) {
```
haxxmaxx, line 227: What is e? error?

haxxmaxx:
Some unit tests would be nice :)

src/utils/actions.js
Caele: true or false, if the reload succeded or not.
I'm not checking for any permissions etc, so if you try to do a reload without update permissions I assume you'll get an generic "Access denied" message as well.

PR #191, opened 2021-03-17
chore: remove flags

Caele:
Some spring cleaning

### DIFF
--- a/src/__tests__/index.spec.js
+++ b/src/__tests__/index.spec.js
@@ -61,7 +61,6 @@ describe('index', () => {
       constraints: 'constraints',
       theme: undefined,
       senseNavigation: 'nav',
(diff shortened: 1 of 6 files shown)


raulvillora:
I guess we should remove the flags from the client as well

raulvillora:
The following two unit tests can be removed since now the flag will be on:

https://github.com/qlik-oss/sn-action-button/blob/master/src/utils/__tests__/navigation-actions.spec.js#L43
https://github.com/qlik-oss/sn-action-button/blob/master/src/utils/__tests__/navigation-actions.spec.js#L48

niekvanstaveren (approved):

raulvillora:
We added a task to remove this flag from the sense-client

PR #197, opened 2021-05-07
chore: husky conf

eliseeborn:
upgrading husky broke the commit hooks


niekvanstaveren (approved):
Looks good if it works I guess 😄

PR #220, opened 2021-09-02
feat: add support for executing automations from button

goldbergjeffrey:
Created new action property panel definition based on Qlik Application Automation.


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.13.0",
   "description": "Supernova action button",
   "license": "MIT",
-  "author": "QlikTech International AB",
+  "author": "Goldberg Productions",
```
eliseeborn, line 6: This seems like an unnecessary change :)

src/index.js
```diff
@@ -35,6 +35,7 @@ export default function supernova(env) {
       }, []);
 
       const layout = useStaleLayout();
+      console.log(layout);
```
eliseeborn, line 38: remember to remove all logs before merging

eliseeborn (commented):
I haven't worked a lot on the actions themselves, so I'll leave reviewing those to Niek or Tobias

package.json
goldbergjeffrey: updated

src/index.js
goldbergjeffrey: done

src/utils/actions.js
```diff
@@ -230,6 +230,77 @@ const actions = [
     },
     requiredInput: ['partial'],
   },
+  {
```
niekvanstaveren, line 233: This action will need to be behind a feature flag. On QSEfW this endpoint will not exist and it should be disabled.

src/utils/actions.js
```diff
@@ -230,6 +230,77 @@ const actions = [
...
+     *
+     * ARGS
+     * app - Reference to current app inherited from index.js
+     * automation - the item id of the automation to contact the items api
+     * automationPostData - boolean value. If true, creates a bookmark of the
+     * current selections and sends the resulting bookmark id as an input
+     * parameter to the selected automation.
+    */
+
+    label: 'Execute Automation',
```
niekvanstaveren, line 247: This will need to use a translation string (similar as the other actions) and that translation string needs to be added to sense-client to be handled in the property panel

src/ext.js
```diff
@@ -153,6 +153,33 @@ export default function ext({ translator }) {
                   defaultValue: false,
                   show: (data) => checkShowAction(data, 'partial'),
                 },
+                // adds automation to actions and adds a dropdown property panel
+                // item to select the automation for the button to trigger
+                automation: {
+                  type: 'string',
+                  component: 'dropdown',
+                  label: 'Select automation',
```
niekvanstaveren, line 161: This needs to be translated

src/ext.js
```diff
@@ -153,6 +153,33 @@ export default function ext({ translator }) {
...
+                  },
+                  show: (data) => checkShowAction(data, 'automation'),
+                },
+                // Boolean property to instruct the automation action to create a
+                // bookmark and send it to the selected automation in the
+                // property panel.
+                automationPostData: {
+                  type: 'boolean',
+                  ref: 'automationPostData',
+                  label: 'Send current selections to automation?',
```
niekvanstaveren, line 179: This would need to be translated. automationPostData and Send current selections to automation does not really make sense together.

src/utils/actions.js
```diff
@@ -230,6 +230,77 @@ const actions = [
...
+    value: 'executeAutomation',
+    getActionCall: ({ app, automation, automationPostData }) => async () => {
+      if (automation !== undefined) {
+        try {
+          const itemInfo = await fetch(`../api/v1/items/${automation}`)
+            .then(response => response.json());
+          const autoInfo = await fetch(`../api/v1/automations/${itemInfo.resourceId}`)
+            .then(response => response.json());
```
niekvanstaveren, line 255: Is the resourceId different from the automation id? Why not store the resourceId from the start. Otherwise you need to retrieve it every time you execute this automation

src/utils/actions.js
```diff
@@ -230,6 +230,77 @@ const actions = [
...
+                for (let i = 0; i < blocks.blocks.length; i++) {
+                  if (blocks.blocks[i].displayName === 'Inputs') {
+                    item = blocks.blocks[i];
+                    break;
+                  }
+                }
+                return item.form[0].label.toLowerCase();
+              });
+            const newDate = new Date();
+            const bmkProp = {
```
niekvanstaveren, line 273: Would these bookmarks show up in the bookmark overview as well? Do they need to be cleaned up in any way?

goldbergjeffrey: Is there a way to create hidden bookmarks still? I'd prefer to do that but if it is possible it's not documented :) Otherwise cleanup would have to happen by the automation if the automation owner adds blocks to remove the bookmark.

src/utils/actions.js
goldbergjeffrey: Yes, the item id is from collections service `${automation}`. The `${itemInfo.resourceId}` is the guid for the automation within automations core service.

src/ext.js
goldbergjeffrey: I think I don't know what I'm doing here because I don't understand what you mean with your comment.

src/ext.js
```diff
@@ -153,6 +153,33 @@ export default function ext({ translator }) {
...
                 },
+                // adds automation to actions and adds a dropdown property panel
+                // item to select the automation for the button to trigger
+                automation: {
+                  type: 'string',
+                  component: 'dropdown',
+                  label: 'Select automation',
+                  ref: 'automation',
+                  options: async () => {
+                    const automations = await fetch('../api/v1/items?resourceType=automation').then((response) =>
```
niekvanstaveren, line 164: We should probably have a look at the usage of fetch. In client we make use of axios. Easy enough to add though. 
The url used should probably be better formed or maybe even as input from the user. This will only work if the button is clicked and configured on the same domain as the automation is running. Pretty sure that is not always the case. 

@Caele Do you know of any restrictions on fetch versus axios? I know that for axios we add the correct headers so probably best to use axios here as well

src/utils/__tests__/actions.spec.js
```diff
@@ -18,6 +22,12 @@ describe('actions', () => {
       bookmark = 'someBookmark';
       variable = 'someVariable';
       qStateName = 'someState';
+      automation = 'someAutomation';
+      executePath = 'someAutomationExecutionPath';
+      automationPostData = false;
+      executeAutomation = {
+        fetch: sinon.spy(),
+      };
```
niekvanstaveren, line 30: This is not the correct way to mock fetch. Let's wait with fixing this until we are sure on what to use for the REST calls

goldbergjeffrey: switched to Axios

src/utils/actions.js
```diff
@@ -8,228 +8,339 @@ export const getValueList = async (app, values, isDate) => {
...
+     * automation - the item id of the automation to contact the items api
+     * automationPostData - boolean value. If true, creates a bookmark of the
+     * current selections and sends the resulting bookmark id as an input
+     * parameter to the selected automation.
+     */
+
+    translation: 'Object.ActionButton.ExecuteAutomation',
+    value: 'executeAutomation',
+    getActionCall:
+      ({ app, automation, automationPostData }) =>
```
Caele, line 289: `app.id` and `automation` comes from properties/layout and cannot be trusted, so they need to be properly escaped when used in URLS and payloads.

niekvanstaveren: Okay. So we should probably add some sanitiser to the button as well. We never concluded that discussion on how to do it right? Copy of a solution or should we start looking into a library that we can use?

goldbergjeffrey: Hi there, let me know how I can help here. I'm ready to jump back in.

goldbergjeffrey:
@niekvanstaveren can you help with coverage? It appears there are still issues with tests?

goldbergjeffrey:
@Caele is encodeURIComponent sufficient? https://github.com/qlik-oss/sn-action-button/commit/399f652484aa2ff7208edf5d81644b3132ea6368

Caele:
> @Caele is encodeURIComponent sufficient? [399f652](https://github.com/qlik-oss/sn-action-button/commit/399f652484aa2ff7208edf5d81644b3132ea6368)

yea I think so

niekvanstaveren (approved):

PR #230, opened 2021-10-08
fix: remove tabindex from actionbutton

AmeliaLinne:
Previously action-button took focus from grid-cell which it shouldnt. This fixes https://jira.qlikdev.com/browse/QB-7434.

Tested on windows and localhost. 

You can still tab to the gridcell the action-button is in and press enter to perform the action.

### DIFF
--- a/src/components/action-button.js
+++ b/src/components/action-button.js
@@ -17,6 +17,7 @@ export default function renderButton({ layout, theme, app, constraints, senseNav
   const isClickable = !disabled && !constraints.active;
   const formattedStyles = styleFormatter.getStyles({ style, disabled, theme, element });
   button.setAttribute('style', formattedStyles);
+  button.setAttribute('tabindex', '-1');
   styleFormatter.createLabelAndIcon({ button, theme, style, isSense });
 
   button.onclick = async () => {


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nebula.js/sn-action-button",
-  "version": "1.14.0",
+  "version": "1.14.1",
```
niekvanstaveren, line 3: Not sure how you updated this, but only changing this number will not do the trick

src/components/action-button.js
```diff
@@ -17,6 +17,7 @@ export default function renderButton({ layout, theme, app, constraints, senseNav
   const isClickable = !disabled && !constraints.active;
   const formattedStyles = styleFormatter.getStyles({ style, disabled, theme, element });
   button.setAttribute('style', formattedStyles);
+  button.setAttribute('tabindex', '-1');
```
niekvanstaveren, line 20: I know this will work, but it feel weird to remove the tabindex from a button. We might instead need to accept that it is not a normal button

package.json
AmeliaAndersson: Is there documentation somewhere on how to do it?

PR #246, opened 2021-11-18
fix: add saveObjects call to bookmark creation on automation execute action

goldbergjeffrey:
Added a doSave to bookmark creation when executing automation with current selections so when app is opened by automation bookmark is available to be applied.

### DIFF
--- a/src/utils/actions.js
+++ b/src/utils/actions.js
@@ -329,7 +329,7 @@ const actions = [
                   .createBookmark(bmkProp)
                   .then((bookmark) => bookmark.getLayout())
                   .then((layout) => layout.qInfo.qId);
-                await app.doSave();
+                await app.SaveObjects();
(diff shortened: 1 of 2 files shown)


src/utils/actions.js
```diff
@@ -329,6 +329,7 @@ const actions = [
                   .createBookmark(bmkProp)
                   .then((bookmark) => bookmark.getLayout())
                   .then((layout) => layout.qInfo.qId);
+                await app.doSave();
```
niekvanstaveren, line 332: You should not do a doSave call here. That is quite heavy on the engine. Better to do a `saveObjects`

goldbergjeffrey: see new commit down below.

niekvanstaveren (approved):

PR #249, opened 2021-11-24
docs: add jsdocs for all properties

haxxmaxx:
- Adding the `spec` command to generate api specs
- Add all properties to `object` properties


spec-configs/props.conf.js
```diff
@@ -0,0 +1,28 @@
...
+  package: path.resolve(__dirname, '../package.json'),
+  api: {
+    stability: 'experimental',
+    properties: {
+      'x-qlik-visibility': 'public',
+    },
+    visibility: 'public',
+    name: `${pkg.name}:properties`,
+    version: pkg.version,
+    description: 'Table generic object definition',
```
cbt1, line 16: ☝️

src/ext.js
```diff
@@ -162,8 +162,9 @@ export default function ext({ translator, automationsEnabled }) {
                   ref: 'automation',
                   dropdownOnly: true,
                   options: async () => {
-                    const automations = await fetch('../api/v1/items?resourceType=automation&limit=100').then((response) =>
-                      response.json());
+                    const automations = await fetch('../api/v1/items?resourceType=automation&limit=100').then(
```
cbt1, line 165: Nitpicking here but I think a common praxis is to line break on new method calls.

```js
const automations = await fetch('../api/v1/items?resourceType=automation&limit=100')
  .then((response) => response.json());
```

src/ext.js
```diff
@@ -260,6 +261,7 @@ export default function ext({ translator, automationsEnabled }) {
               component: 'switch',
               translation: 'properties.enableToggle',
               ref: 'useEnabledCondition',
+              defaultValue: false,
```
cbt1, line 264: Is there a risk that these default values have changed now that they are explicitly set?

api-specifications/properties.json
```diff
@@ -0,0 +1,625 @@
+{
+  "scriptappy": "1.0.0",
+  "info": {
+    "name": "@nebula.js/sn-action-button:properties",
+    "description": "Table generic object definition",
```
cbt1, line 5: ☝️

api-specifications/properties.json
```diff
@@ -0,0 +1,625 @@
...
+    "version": "1.14.3",
+    "license": "MIT",
+    "stability": "experimental",
```
cbt1, line 8: Is this still experimental?

cbt1:
Ideally this PR should be split into a `docs` and `refactor` PR. As you are now making code changes and updating the docs, so your `docs` tag is not really valid.

src/ext.js
haxxmaxx: well yes, if I did it incorrectly. I can double-check, maybe even do some manual sanity checks

src/ext.js
haxxmaxx: sure, lint apparently changed this

haxxmaxx:
> Ideally this PR should be split into a `docs` and `refactor` PR. As you are now making code changes and updating the docs, so your `docs` tag is not really valid.

Hmm good point. I guess I could start with the code changes

api-specifications/properties.json
haxxmaxx: Not sure, when does it stop being experimental?

src/object-properties.js
```diff
@@ -1,38 +1,62 @@
...
+
 const properties = {
   /**
-   * @type {boolean}
+   * Current version of this generic object definition.
+   * @type {string}
+   * @default
    */
-  showTitles: false,
+  version: (process && process.env && process.env.PACKAGE_VERSION) || '1.0.0',
```
cbt1, line 14: When is the version nbr not available?

haxxmaxx: When running rendering tests, but it is still failing. process comes from node it seems, I'll have to get around it in a different way

cbt1 (approved):
LGTM, I guess you need to figure out why the build is failing

PR #250, opened 2021-11-25
chore: update build scripts

haxxmaxx:
This enables some useful steps for building and eventually publishing the package. There will be another PR with some changes to the CI config.


tools/validate-nebula-package.js
```diff
@@ -0,0 +1,115 @@
...
+};
+
+// known charts that are native (build time) in Sense
+const validate = (pkg, dir) => {
+  validatePackageJsonContent(pkg);
+  validateFiles(pkg);
+
+  const cleanedPkg = JSON.stringify(pkg, null, 2);
+  // package version must be 0.x.x at the moment
+  if (!/1\.\d\.\d+/.test(pkg.version)) {
```
cbt1, line 90: Care to elaborate on the purpose here? So you are not allowed to have any other version than 1.x.x? Even if the error msg and comment says something else. So version 2.x.x should not be allowed?

haxxmaxx: Not sure, it is copy pasted from table. i assume it is blocking doing `major` version bumps. And then when it was actually bumped the regex was changed. I'll look closer at it

cbt1: Ideally, IMO, version should be based on conventional commits. In which case there are other toolsets to deal with making sure the correct version is set.

cbt1: Other than that, I'm not familiar with these reasoning behind these scripts but I see that they have already been approved for the table, so I see no reason to not allow them here as well.

cbt1 (approved):

tools/validate-nebula-package.js
haxxmaxx: Yes, I know. I think that was Miralems suggestion back when I did the KPI. Didn't have enough knowledge/time to do it then, kind of the same now. I think we should come up with a good way to release the charts that are in seperate repos, and then use that for all charts for consistency

cbt1: I'm kind of looking at that now actually, tho it will be a simple processes initially.

LiKang6688: the comment is wrong.  should be // package version must be 1.x.x at the moment

haxxmaxx: I just changed it to a more general check, so that it is only digits divided with dots. Just to catch if it is on the wrong format

LiKang6688 (approved):

PR #251, opened 2021-11-25
refactor: move properties defaults to ext

haxxmaxx:
Moving the default values for styling to `ext` and doing sanity checking for all options in `style-formatter`. This enables mashup builders to set only the options they want to change.


yarn.lock
```diff
@@ -8651,16 +8651,16 @@ picomatch@^2.0.7:
   resolved "https://registry.yarnpkg.com/picomatch/-/picomatch-2.2.1.tgz#21bac888b6ed8601f831ce7816e335bc779f0a4a"
   integrity sha512-ISBaA8xQNmwELC7eOjqFKMESB2VIqt4PPDD0nsS95b/9dZXvVKOlz9keMSnoGGKcOHXfTvDD6WMaRoSc9UuhRA==
 
-picomatch@^2.2.1, picomatch@^2.2.2:
-  version "2.2.2"
-  resolved "https://registry.yarnpkg.com/picomatch/-/picomatch-2.2.2.tgz#21f333e9b6b8eaff02468f5146ea406d345f4dad"
-  integrity sha512-q0M/9eZHzmr0AulXyPwNfZjtwZ/RBZlbN3K3CErVrk50T2ASYI7Bye0EvekFY3IP1Nt2DHu0re+V2ZHIpMkuWg==
-
-picomatch@^2.2.3:
+picomatch@^2.2.1, picomatch@^2.2.3:
```
cbt1, line 8654: Unintentional yarn update?

src/utils/style-formatter.js
```diff
@@ -38,16 +38,16 @@ const getColor = ({ useColorExpression, colorExpression, color }, defaultColor,
...
-    const { font, background, border } = style;
+    const { font = {}, background = {}, border = {} } = style;
     const primaryColor = theme.getDataColorSpecials().primary;
     // enable
     styles += disabled ? formatProperty('opacity', 0.4) : formatProperty('cursor', 'pointer');
     // font
     styles += formatProperty('color', getColor(font, '#ffffff', theme));
-    font.style.bold && (styles += formatProperty('font-weight', 'bold'));
-    font.style.italic && (styles += formatProperty('font-style', 'italic'));
+    font.style && font.style.bold && (styles += formatProperty('font-weight', 'bold'));
```
cbt1, line 49: Seems to me like the default value for `bold` has changed here? Since font style will now be an empty object right? Was that intentional?

haxxmaxx: You're right. nice find. I'm working on putting all defaults in a separate folder so style-formatter and ext always map correctly. I can ping you when I'm done so you don't have to review more than necessary

haxxmaxx: You're right. nice find. I'm working on putting all defaults in a separate folder so style-formatter and ext always map correctly. I can ping you when I'm done so you don't have to review more than necessary

cbt1: Great as my only concern here would be that the default values have changed in the mashup case. Which I believe would be a breaking change, unless the API is experimental.

src/utils/style-formatter.js
```diff
@@ -46,8 +51,9 @@ export default {
     styles += disabled ? formatProperty('opacity', 0.4) : formatProperty('cursor', 'pointer');
     // font
     styles += formatProperty('color', getColor(font, '#ffffff', theme));
-    font.style && font.style.bold && (styles += formatProperty('font-weight', 'bold'));
-    font.style && font.style.italic && (styles += formatProperty('font-style', 'italic'));
+    const fontStyle = font.style || DEFAULTS.FONT_STYLE;
```
cbt1, line 54: Where's `underline`?

yarn.lock
haxxmaxx: Don't think so, I can reinstall it

src/utils/style-formatter.js
haxxmaxx: fyi the api hasn't been published yet so we are good on that front. But it should correspond to creating a default button in the client

src/utils/style-formatter.js
haxxmaxx: It is applied in `createLabelAndIcon`. Don't remember why, but it was something about the scaling of the label I think

cbt1 (approved):

PR #253, opened 2021-11-25
feat: make rendering testing work with enigma-mocker and test fixtures

LiKang6688:
Nebula serve using mocked Qlik core engine (Enigma mocker) to render the visualization charts based on the testing fixture file

Testing cases using Puppeteer capture screenshots of those rendered charts and matches against the baseline to ensure our charts look as intended and present clients with expected visuals.

Visual regression Testing flow:

1. Start Nebula serve and launch Puppeteer
2. Iterate testing fixture files
3. Create test case per testing fixture file
4. Render chart based on testing fixture file in Nebula serve using Enigma mocker
5. Puppeteer capture screenshot
6. Compare screenshot with the baseline image


test/rendering/__fixtures__/scenario_1.fix.js
```diff
@@ -0,0 +1,102 @@
+export default () => ({
```
haxxmaxx, line 1: Now the logs on circle say `scenario_1`, should have a more descriptive name.

.circleci/config.yml
```diff
@@ -55,17 +46,11 @@ jobs:
...
           environment:
             JEST_JUNIT_OUTPUT_DIR: ./coverage/junit/
-
       - run:
           name: Integration tests
-          command: |
-            DocId=$(curl -s --data-binary @data/apps/Executive_Dashboard.qvf http://localhost:9076/v1/apps/import | jq -r '.attributes.id')
-            APP_ID=$DocId yarn run test:integration --chrome.browserWSEndpoint "ws://localhost:3000" --no-launch
-
+          command: yarn test:rendering
```
haxxmaxx, line 51: Do we want to call it integration tests or rendering tests? If we add actual integration tests, should they be a separate step? If so I think this should be called Rendering tests

LiKang6688: Right. gonna fix. thanks.

test/rendering/__fixtures__/scenario_1.fix.js
LiKang6688: Yes, I agree. when we go through each feature to create a fixture file, we should have a more descriptive name. Not this time though, this fixture file which is randomly created will be removed in the future anyway.

.circleci/config.yml
LiKang6688: Done

test/rendering/__fixtures__/scenario_1.fix.js
haxxmaxx: ok fine

haxxmaxx (approved):
Same thing here. Make sure you test it in the client since we bumped the node version :)

PR #258, opened 2021-12-16
feat: refresh dynamics and open odag popup

ymchandra:
- refresh dynamic views action
 - open odag popup navigation
 - show odag links for navigation


src/components/action-button.js
```diff
@@ -26,15 +30,20 @@ export const renderButton = ({ layout, theme, app, constraints, senseNavigation,
...
       actions.forEach((action) => {
         const actionObj = allActions.find((act) => act.value === action.actionType);
-        actionObj && actionCallList.push(actionObj.getActionCall({ app, qStateName, ...action }));
+        if (actionObj) {
+          if (actionObj.group === 'dynamicViews') {
+            actionCallList.push(actionObj.getActionCall({ app, senseNavigation }));
+          } else {
+            actionCallList.push(actionObj.getActionCall({ app, qStateName, ...action }));
+          }
+        }
```
niekvanstaveren, line 39: No need for splitting here. Just add senseNavigation as input. The function can decide what to use from the input

src/utils/actions.js
```diff
@@ -342,9 +342,26 @@ const actions = [
...
+          if (typeof senseNavigation.refreshDynamicViews === 'function') {
+            await senseNavigation.refreshDynamicViews();
+          }
+        },
+    requiredInput: [],
+    featureFlag: 'REFRESH_DYNAMIC_VIEWS_ODAG_POPUP',
   },
 ];
 
+export const getFFEnabledOrDefaultActions = (isEnabled) => actions.filter((a) => !a.featureFlag || isEnabled(a.featureFlag));
```
niekvanstaveren, line 363: This name is very confusing. Can it just be getActionsList or something?

src/utils/navigation-actions.js
```diff
@@ -104,8 +104,22 @@ const navigationActions = [
...
+        if (typeof senseNavigation.openOdagPopup === 'function' && odagLink) {
+          await senseNavigation.openOdagPopup(app, odagLink, element);
+        }
+      },
+    requiredInput: ['odagLink'],
+    featureFlag: 'REFRESH_DYNAMIC_VIEWS_ODAG_POPUP',
+  },
 ];
 
+export const getFFEnabledOrDefaultNavigations = (isEnabled) => navigationActions.filter((n) => !n.featureFlag || isEnabled(n.featureFlag));
```
niekvanstaveren, line 121: Also a bit of a confusing name.

src/utils/navigation-actions.js
```diff
@@ -104,8 +104,22 @@ const navigationActions = [
     },
     requiredInput: ['websiteUrl'],
   },
+  {
+    translation: 'Object.ActionButton.SelectOdagApp',
+    value: 'odagLink',
+    navigationCall:
+      async ({ app, senseNavigation, odagLink, element }) => {
+        if (typeof senseNavigation.openOdagPopup === 'function' && odagLink) {
+          await senseNavigation.openOdagPopup(app, odagLink, element);
```
niekvanstaveren, line 113: How will this work outside the client? E.g. embedded sheet, embedded button, mashups...

src/utils/__tests__/navigation-actions.spec.js
```diff
@@ -118,6 +119,27 @@ describe('navigation actions', () => {
         expect(senseNavigation.goToSheet).toHaveBeenCalledWith('id3');
       });
     });
+
+    describe('odag navigation popup', () => {
+      it('should not call openOdagPopup when openOdagPopup is undefined', async () => {
+        const navigationObject = navigationActions.find((navigation) => navigation.value === 'odagLink');
```
niekvanstaveren, line 125: I think you would only need to do this once and use it in al three test cases

src/components/action-button.js
```diff
@@ -13,7 +13,11 @@ export const renderButton = ({ layout, theme, app, constraints, senseNavigation,
   const isSense = !!senseNavigation;
   const button = element.firstElementChild;
   const { style, qStateName } = layout;
-  const disabled = layout.useEnabledCondition && layout.enabledCondition === 0;
+  const { navigation } = layout;
+  const { odagLink } = layout;
+  let disabled = layout.useEnabledCondition && layout.enabledCondition === 0;
+  const isOdagLink = navigation.action === 'odagLink' && odagLink.length === 0;
+  disabled = isOdagLink || disabled;
```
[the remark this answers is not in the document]
ymchandra: It will have odagLink as empty string and when the user clicks on that, it sends empty value to the Odag backend. This is resulting in errors. That's why I am disabling it.

src/components/action-button.js
ymchandra: Done.

src/utils/actions.js
ymchandra: Now the name is changed.

src/utils/navigation-actions.js
ymchandra: Now the name is changed.

src/utils/navigation-actions.js
ymchandra: I don't think the Odag will work outside the sense-client now. So this will not work either. This change is specifically to create Odag from button object.

src/utils/__tests__/navigation-actions.spec.js
ymchandra: Done.

src/utils/navigation-actions.js
niekvanstaveren: Understand that, but you can create a button in sense client (and configure) and then have the button outside by embedding a sheet for instance. You need to test these scenarios

src/components/action-button.js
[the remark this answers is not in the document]
niekvanstaveren: You also do a check on odagLink in the navigation call itself. That should be enough. Now you are also blocking other actions from happening.

src/components/action-button.js
```diff
@@ -13,7 +13,11 @@ export const renderButton = ({ layout, theme, app, constraints, senseNavigation,
   const isSense = !!senseNavigation;
   const button = element.firstElementChild;
   const { style, qStateName } = layout;
-  const disabled = layout.useEnabledCondition && layout.enabledCondition === 0;
+  const { navigation } = layout;
+  const { odagLink } = layout;
```
niekvanstaveren, line 17: These can be in one line.

src/components/action-button.js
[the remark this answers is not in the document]
ymchandra: Ok, now the button is not disabled, instead the navigation action is restricted based on odagLink value.

src/utils/navigation-actions.js
ymchandra: In embedded sheet mode, the button is trying to make backend call and couldn't fetch data and failing there which is expected.

src/components/action-button.js
ymchandra: Done

src/components/action-button.js
```diff
@@ -13,6 +13,7 @@ export const renderButton = ({ layout, theme, app, constraints, senseNavigation,
   const isSense = !!senseNavigation;
   const button = element.firstElementChild;
   const { style, qStateName } = layout;
+  const { navigation, odagLink } = layout;
```
niekvanstaveren, line 16:
```suggestion
  const { style, qStateName, navigation, odagLink } = layout;
```

src/ext.js
```diff
@@ -248,6 +248,22 @@ export default function ext({ translator, automationsEnabled }) {
                   show: (data) => checkShowNavigation(data, 'websiteUrl'),
                   defaultValue: false,
                 },
+                odagLink: {
+                  type: 'string',
+                  ref: 'odagLink',
```
niekvanstaveren, line 253: Why are you not storing this on `navigation.odagLink` ?

src/components/action-button.js
ymchandra: Done.

src/ext.js
ymchandra: Done.

src/utils/actions.js
```diff
@@ -342,9 +342,26 @@ const actions = [
...
+    group: 'dynamicViews',
+    getActionCall:
+      ({ senseNavigation }) =>
+        async () => {
+          if (typeof senseNavigation.refreshDynamicViews === 'function') {
+            await senseNavigation.refreshDynamicViews();
+          }
+        },
+    requiredInput: [],
+    featureFlag: 'REFRESH_DYNAMIC_VIEWS_ODAG_POPUP',
```
niekvanstaveren, line 359: Like this btw. Great work! 😄

niekvanstaveren (approved):

PR #260, opened 2021-12-16
build: add the extension build to the package

haxxmaxx:
Using the same script as scatter plot


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@nebula.js/sn-action-button",
   "version": "1.14.3",
-  "description": "Supernova action button",
+  "description": "Button chart for executing actions",
```
LiKang6688, line 4: is it a chart really 🤔

package.json
```diff
@@ -22,17 +22,18 @@
...
   "engines": {
     "node": ">=16"
   },
   "main": "dist/sn-action-button.js",
   "module": "dist/sn-action-button.esm.js",
   "scripts": {
-    "build": "node ./tools/build.js --core && shx cp assets/* dist",
-    "build:dev": "node ./tools/build.js --core --mode development && shx cp assets/* dist",
-    "build:watch": "node ./tools/build.js --mode development -w",
+    "build": "node ./tools/build.js --core --ext && shx cp assets/* dist",
```
LiKang6688, line 34: the files in assets can also be generated by nebula sense, right? So do we still need assets?

haxxmaxx: there is an example picture in there, i think it still should be there

LiKang6688: right, preview image can not be generated by sense.

package.json
haxxmaxx: you find it under charts in the client, but it is a bit weird I agree. I change it
Screenshot 2021-12-20 at 11 19 22
.

LiKang6688: Good that you bring it up. As I understand that the button in the Charts is not sn-action-button yet, so I found that in the assets folder, the preview.gift does not look right to me. It should be a static image that just looks like in your attached image, the icon image which is left to the Button.

package.json
LiKang6688: I had a try in this branch.

Screenshot 2021-12-20 at 14 54 44
the preview.gif is not copied into sn-action-button-ext. Not sure why. I do see preview.gif in assets and dist folders.

package.json
LiKang6688: aha, I see. Thank you! Then I believe that the gif should be moved to other places for readme, and not in the assets folder which results in being copied into the dist folder to increase the size for nothing.

haxxmaxx: Well all the other repos have images stored in the assets folder. I think the qext file ended up in there and thus the whole folder was copied. But as you wrote in the other comment, unclear if we still need to copy assets. I think we should just not copy it. No reason to have images in the package, just extra size. There are plenty of documentation on github and dev portal

LiKang6688 (approved):

PR #263, opened 2022-01-11
ci: update script to publish from ci

haxxmaxx:
Based on sn-tables ci script. Publishes if on master, the tag and version in package.json has been bumped

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,9 @@
 version: 2
 
+defaults: &defaults
+  working_directory: ~/action-button
+  # Available images https://hub.docker.com/r/circleci/node/tags/
+
 aliases:
   - &restore_yarn_cache
     name: Restore yarn cache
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -53,4 +75,96 @@ jobs:
...
           path: coverage/junit
-      - save_cache: *save_yarn_cache
+      - persist_to_workspace:
+          root: ~/action-button
+          paths:
+            - dist
+            - package.json
+            - core
+
+  # publish-dev:
```
haxxmaxx, line 85: I've never used this, but we have it for several charts. I guess it can be removed?

niekvanstaveren: I guess you can use it to test your extension? Is it commented in all other charts as well? In that case better to remove I think

niekvanstaveren (commented):
Looks good. Should check a release after this probably

niekvanstaveren:
Probably you want to update the README as well

.circleci/config.yml
haxxmaxx: No I commented it now

haxxmaxx:
> Probably you want to update the README as well

Ah yes, will fix that

.github/CONTRIBUTION.md
```diff
@@ -0,0 +1,53 @@
+## Developing
+
+1. Install with `yarn`
+1. Transpile code: `yarn build` (or `yarn build:watch`)
+1. Link to sense-client (locally on your computer):
```
haxxmaxx, line 5: Not sure we want this step here, you need access to sense-client. @niekvanstaveren

niekvanstaveren: No we should not have this in the open source repo. Probably only mention yarn start here.

haxxmaxx: done

niekvanstaveren (approved):

PR #268, opened 2022-01-20
fix: changed hasInputs to inputs

goldbergjeffrey:
Needed to update hasInputs to inputs to support change in automation api.

### DIFF
--- a/src/utils/__tests__/actions.spec.js
+++ b/src/utils/__tests__/actions.spec.js
@@ -11,7 +11,7 @@ describe('actions', () => {
   let bookmark;
   let variable;
   let automation;
-  let hasinputs = false;
+  let inputs = false;
   let automationPostData = false;
   const value = 'someValue';
   const softLock = true;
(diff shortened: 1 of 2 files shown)


niekvanstaveren (approved):

src/utils/__tests__/actions.spec.js
```diff
@@ -306,7 +306,7 @@ describe('actions', () => {
     });
 
     it('should call executeAutomation', async () => {
-      hasinputs = false;
+      inputs = [];
```
niekvanstaveren, line 309: This should not be necessary

src/utils/__tests__/actions.spec.js
```diff
@@ -326,7 +326,7 @@ describe('actions', () => {
     });
 
     it('should call executeAutomation with creation of bookmark and save app', async () => {
-      hasinputs = true;
+      inputs = [];
```
niekvanstaveren, line 329: Don't you want to test a flow with inputs here?

niekvanstaveren (approved):

PR #272, opened 2022-02-02
chore: version not being passed to variable

joakim-ronning:
attempting to fix version not being passed to the VER variable correctly on circle ci.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,6 +149,7 @@ jobs:
           name: Run API Compliance
           command: >
             VER=$(cat ./version.txt)
+
             docker run --volumes-from specs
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -149,6 +149,7 @@ jobs:
           name: Run API Compliance
           command: >
             VER=$(cat ./version.txt)
+
```
haxxmaxx, line 152: does this make a difference?

joakim-ronning: Yes. folded block style (https://yaml.org/spec/1.2-old/spec.html#id2796251) in yaml treats consecutive lines as a single line. In order to treat these commands as separate commands, they must be separated by a blank line.

haxxmaxx: ok, good to know :)

haxxmaxx (approved):

PR #276, opened 2022-03-02
feat: document chaining navigation option

niekvanstaveren:
Allowing the document chain navigation action:
image

Using the temporary bookmark we can apply selections from one app in a different app. 

Also updating a lot of assertions since they were not doing anything.


src/ext.js
```diff
@@ -196,6 +196,13 @@ export default function ext({ translator, isEnabled, senseNavigation }) {
                   options: getNavigationsList(isEnabled),
                   dropdownOnly: true,
                 },
+                appId: {
```
haxxmaxx, line 199: This needs to be added to the api

src/utils/navigation-actions.js
```diff
@@ -102,23 +102,40 @@ const navigationActions = [
         // no-op
       }
     },
-    requiredInput: ['websiteUrl'],
+    requiredInput: ['websiteUrl', 'sameWindow'],
+  },
+  {
+    translation: 'Object.ActionButton.DocumentChain',
+    value: 'openChainedApp',
```
haxxmaxx, line 109: This also needs to be in the api

haxxmaxx (approved):
lgtm, not sure what is going on with the api-gpvernance

PR #278, opened 2022-03-07
docs: added api specs for refresh dynamic views and odag launch

ymchandra:
Updated API spec changes for refresh dynamic views and odag launch changes.

### DIFF
--- a/src/object-properties.js
+++ b/src/object-properties.js
@@ -120,7 +120,7 @@ const properties = {
  * @property {string=} story - Story ID. Required for 'goToStory'
  * @property {string=} websiteUrl - URL for website. required for 'openWebsite'
  * @property {boolean=} sameWindow - Set to true to open in same window/tab. Required for 'openWebsite'
- * @property {string=} odagLinkName - odagLink name. Required for 'odagLink'
+ * @property {string=} odagLink - ODAG link name. Required for 'odagLink'
  */
 
 /**
--- a/src/object-properties.js
+++ b/src/object-properties.js
@@ -101,7 +101,7 @@ const properties = {
  * Defines what action to perform and options for that action type.
  * @typedef {object} Action
  * @property {string} [actionLabel=''] - Label for reference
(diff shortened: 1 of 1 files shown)


src/object-properties.js
```diff
@@ -115,11 +115,12 @@ const properties = {
 /**
  * Defines a navigation action to perform after the other actions. Note that when the sn-action-button is used outside Qlik Sense you can only use the 'openWebsite' action.
  * @typedef {object} NavigationAction
- * @property {'nextSheet'|'prevSheet'|'firstSheet'|'lastSheet'|'goToSheet'|'goToSheetById'|'goToSheetById'|'goToStory'|'openWebsite'} action - Name of navigation action
+ * @property {'nextSheet'|'prevSheet'|'firstSheet'|'lastSheet'|'goToSheet'|'goToSheetById'|'goToSheetById'|'goToStory'|'openWebsite'|'odagLink'} action - Name of navigation action
  * @property {string=} sheet - sheet ID. Required for 'goToSheet' and 'goToSheetById'
  * @property {string=} story - Story ID. Required for 'goToStory'
  * @property {string=} websiteUrl - URL for website. required for 'openWebsite'
  * @property {boolean=} sameWindow - Set to true to open in same window/tab. Required for 'openWebsite'
+ * @property {string=} odagLinkName - odagLink name. Required for 'odagLink'
```
haxxmaxx, line 123: Had a look in the code and this isn't what the property is called. But it should probably be called that

haxxmaxx: The API should contain the exact name of the properties. I realized you probably can't change it, so then it should be only `odagLink`.

haxxmaxx: Then in the description you can describe that it is the name
```suggestion
 * @property {string=} odagLink - ODAG link name. Required for 'odagLink'
```

ymchandra: Done!

haxxmaxx (approved):

PR #279, opened 2022-03-16
chore: update machine image

Caele:
Update machine image as the old is getting deprecated.


niekvanstaveren (approved):
2004 current? 😄

Caele:
2004 = 20.04 =)

PR #286, opened 2022-03-23
fix: center button within its parent element

LiKang6688:


niekvanstaveren:
1.14 is already done here right? https://github.com/qlik-oss/sn-action-button/pull/284

PR #293, opened 2022-04-25
feat: disable several features on direct query mode

LiKang6688:
disable bookmarks, document chaining, go to a story, and reload data on direct query mode


niekvanstaveren (changes requested):
Why not the show function as we talked about? The input to the show function can be isEnabled, isFeatureBlacklisted and isUnsupportedFeature

LiKang6688:
> Why not the show function as we talked about? The input to the show function can be isEnabled, isFeatureBlacklisted and isUnsupportedFeature

In that case, I need to add `show: boolean | function` to every item within `actions` and `navigationActions`.
I think this solution is adding less code and easier to read.

cbt1:
> 

I don't think your currently solution is wrong as a `show` callback would require slightly more code, but you could reduce the code complexity as follows:

```js
// Current
export const getActionsList = (isEnabled, isHidden) => actions.filter((a) => (!a.featureFlag || isEnabled(a.featureFlag)) && !a.hide?.(isHidden));

// Or with only a hide callback, where the items with a feature flag would also need to be updated with a "hide callback"
export const getActionsList = (isEnabled, shouldHide) =>  actions.filter((a) => !a.hide?.(isEnabled, shouldHide));
```

LiKang6688:
> > 
> 
> I don't think your currently solution is wrong as a `show` callback would require slightly more code, but you could reduce the code complexity as follows:
> 
> ```js
> // Current
> export const getActionsList = (isEnabled, isHidden) => actions.filter((a) => (!a.featureFlag || isEnabled(a.featureFlag)) && !a.hide?.(isHidden));
> 
> // Or with only a hide callback, where the items with a feature flag would also need to be updated with a "hide callback"
> export const getActionsList = (isEnabled, shouldHide) =>  actions.filter((a) => !a.hide?.(isEnabled, shouldHide));
> ```

Cool. Work on it.

LiKang6688:
@niekvanstaveren @cbt1 Please have an another look.

src/index.js
```diff
@@ -20,10 +20,16 @@ export default function supernova(env) {
     sense,
     translator,
     flags: { isEnabled },
+    anything,
   } = env;
-  const senseNavigation = sense && sense.navigation;
+  const senseNavigation = sense?.navigation;
+  const isFeatureBlacklisted = anything.sense?.isFeatureBlacklisted;
+  const isUnsupportedFeature = anything.sense?.isUnsupportedFeature;
+  const shouldHide = (feature) =>
```
cbt1, line 28: I don't think it's a good idea to batch the hide condition like this. There is a risk, albeit a small risk, that this can produce a false positive. For example if a blacklisted feature is check with `isFeatureBlacklisted('bookmarks')` and returns false but `isUnsupportedFeature('bookmarks')` return true.

src/utils/actions.js
```diff
@@ -347,7 +349,7 @@ const actions = [
         }
       },
     requiredInput: ['automation'],
-    featureFlag: 'ACTION_BUTTON_AUTOMATIONS',
+    hide: (shouldHide) => shouldHide?.('ACTION_BUTTON_AUTOMATIONS'),
```
cbt1, line 352: A possible fix to my other comment is to call `hide` with all three methods.

Ex: `hide: ({ isEnabled }) => !isEnabled('ACTION_BUTTON_AUTOMATIONS')`

cbt1 (changes requested):

src/index.js
LiKang6688: I do not see that it produces a false positive. In your example, the shouldHide is true which looks correct to me. Did I miss anything?

cbt1: My assessment was based off that it should only hide given a specific scenario. So for example the apply bookmarks actions should ONLY hide given it's a blacklisted feature. So it should not matter if it's a disabled feature flag or unsupported feature.

Does that make sense?

LiKang6688: Make sense. Thx!

src/utils/actions.js
LiKang6688: Fix.

cbt1 (approved):

niekvanstaveren (approved):
The name shouldHide might be a bit confusing, but I also don't have something better for it right now

PR #294, opened 2022-04-26
test: migrate from puppeteer to playwright

LiKang6688:


.circleci/config.yml
```diff
@@ -68,13 +68,18 @@ jobs:
             fi
           environment:
             JEST_JUNIT_OUTPUT_DIR: ./coverage/junit/
+      - run:
+          name: Install playwright
+          command: npx playwright install
```
niekvanstaveren, line 73: Why are we using npx to install this? Can we have it through npm dependencies instead?

.github/CONTRIBUTION.md
```diff
@@ -8,15 +8,24 @@
...
 
-`yarn test:rendering`
+    # Install dependencies
+    yarn --frozen-lockfile
+
+    # Build nebula.js visualization
+    yarn build
+
+    chmod 777 ./test/rendering/commands/run-rendering-test.sh
+    yarn test:local:rendering
```
niekvanstaveren, line 26: Should this be one copy block?

.circleci/config.yml
LiKang6688: the name is misleading which is from their docs. 
it is actually installing supported browsers.
I will fix it to install supported browsers.
https://playwright.dev/docs/ci

.github/CONTRIBUTION.md
LiKang6688: Screenshot 2022-05-03 at 11 50 08
it is one copy block. Do I understand you correctly?

niekvanstaveren (approved):

PR #312, opened 2022-08-25
feat: adding sizing options to label

niekvanstaveren:
Adding new sizing options to the button:
- Fluid (relative). The label font size will be calculated based on the box size of the button regardless of the actual label. This means that buttons that have the same size always have the same font size
- Fixed. A fixed size by pixel

### DIFF
--- a/src/ext.js
+++ b/src/ext.js
@@ -26,7 +26,23 @@ const toggleOptions = [
   },
 ];
 
+const fontSizeOptions = [
+  {
+    value: 'responsive',
+    translation: 'properties.responsive',
+  },
+  {
+    value: 'relative',
+    translation: 'properties.fluid',
+  },
+  {
+    value: 'fixed',
+    translation: 'properties.fixed',
+  },
+];
+
 export default function ext({ translator, shouldHide, senseNavigation }) {
+  const { isEnabled } = shouldHide;
   return {
     definition: {
       type: 'items',
@@ -333,6 +349,14 @@ export default function ext({ translator, shouldHide, senseNavigation }) {
                 sizeAndColor: {
                   type: 'items',
                   items: {
+                    fontSizeBehavior: {
+                      component: 'dropdown',
+                      ref: 'style.font.sizeBehavior',
+                      options: fontSizeOptions,
+                      translation: 'properties.kpi.layoutBehavior',
+                      defaultValue: 'responsive',
+                      show: () => isEnabled('IM_324_ACTION_BUTTON_SIZES'),
+                    },
                     fontSize: {
                       component: 'slider',
                       type: 'number',
@@ -341,6 +365,19 @@ export default function ext({ translator, shouldHide, senseNavigation }) {
                       min: 0.2,
                       max: 1,
                       step: 0.01,
+                      show: (data) =>
+                        !isEnabled('IM_324_ACTION_BUTTON_SIZES') ||
+                        !data.style.font.sizeBehavior ||
+                        data.style.font.sizeBehavior !== 'fixed',
+                    },
+                    fontSizeFixed: {
+                      type: 'number',
+                      ref: 'style.font.sizeFixed',
+                      expression: 'optional',
+                      min: 5,
+                      defaultValue: 20,
+                      show: (data) =>
+                        isEnabled('IM_324_ACTION_BUTTON_SIZES') && data.style.font.sizeBehavior === 'fixed',
                     },
                     useFontColorExpression: {
                       ref: 'style.font.useColorExpression',
--- a/src/object-properties.js
+++ b/src/object-properties.js
@@ -65,8 +65,10 @@ const properties = {
       color: DEFAULTS.COLOR,
       colorExpression: '',
       size: DEFAULTS.FONT_SIZE,
+      sizeFixed: DEFAULTS.FONT_SIZE_FIXED,
       style: DEFAULTS.FONT_STYLE,
       align: DEFAULTS.TEXT_ALIGN,
+      sizeBehavior: DEFAULTS.SIZE_BEHAVIOR,
     },
     background: {
       useColorExpression: false,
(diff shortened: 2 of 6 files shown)


src/utils/style-formatter.js
```diff
@@ -91,12 +99,12 @@ export default {
     const { icon = {}, font = {}, label = DEFAULTS.LABEL } = style;
     // text element wrapping label and icon
     const text = document.createElement('text');
-    text.style.whiteSpace = 'nowrap';
+    text.style.whiteSpace = 'pre';
```
niekvanstaveren, line 102: This allows for multiple lines in a label

src/object-properties.js
```diff
@@ -140,11 +142,13 @@ const properties = {
  * @property {PaletteColor=} color - Color defined by index or hex code, needed if useColorExpression is false
  * @property {ColorExpression=} colorExpression - Color defined by expression, needed if useColorExpression is true
  * @property {number=} [size=0.5] - Relative[] size of label, must be greater than 0 and 1 fills the entire button
+ * @property {number=} [sizeFixed=20] - Fixed size of the label in pixels. Must be bigger than 5. Only active when sizeBehavior is fixed.
```
niekvanstaveren, line 145: @haxxmaxx The api specifications would be updated when actually releasing right?

src/utils/style-formatter.js
```diff
@@ -42,6 +42,14 @@ const getColor = (
   return !resolvedColor || resolvedColor === 'none' ? defaultColor : resolvedColor;
 };
 
+export function measureText(text, font) {
+  const canvas = measureText.canvas || (measureText.canvas = document.createElement('canvas'));
+  const context = canvas.getContext('2d');
+  context.font = font;
+  const metrics = context.measureText(text);
+  return metrics;
+}
```
niekvanstaveren, line 51: Not really sure we want this. It will allow in the feature to calculate this with different fonts, but right we are only able to use one specific font here.

src/ext.js
```diff
@@ -26,7 +26,23 @@ const toggleOptions = [
...
+    translation: 'properties.fluid',
+  },
+  {
+    value: 'fixed',
+    translation: 'properties.fixed',
+  },
+];
+
 export default function ext({ translator, shouldHide, senseNavigation }) {
+  const { isEnabled } = shouldHide;
```
eliseeborn, line 45: what is shouldHide? is that just that we don't want feature flag handling to be visible in open source projects?

niekvanstaveren: No shouldHide actually holds a couple of functions. From index.js:

```const shouldHide = { isEnabled, isFeatureBlacklisted, isUnsupportedFeature };```

File: src/ext.js [hunk 5]
```diff
@@ -336,6 +352,14 @@ export default function ext({ translator, shouldHide, senseNavigation }) {
                 sizeAndColor: {
                   type: 'items',
                   items: {
+                    fontSizeBehavior: {
+                      component: 'dropdown',
+                      ref: 'style.font.sizeBehavior',
+                      options: fontSizeOptions,
+                      translation: 'properties.kpi.layoutBehavior',
```
niekvanstaveren, line 359: A bit ugly because it uses an KPI translation. It probably should not have been a specific KPI translation from the start.

src/object-properties.js
haxxmaxx: that is the way we do it for the table, guessing same here. The API governance will complain if the api specs don't match with the specs for the current version I think

src/ext.js
```diff
@@ -26,7 +26,23 @@ const toggleOptions = [
   },
 ];
 
+const fontSizeOptions = [
+  {
+    value: 'responsive',
+    translation: 'properties.responsive',
+  },
+  {
+    value: 'relative',
```
haxxmaxx, line 35: What is the difference between these?

niekvanstaveren: Responsive will always fit the text inside the button, so it considers the length of the label. 
Relative used the box size of the button to calculate the font size. It is independent of the label (length). This means that buttons with the same size will always have the same font size
Fixed is a fixed size in pixels

src/object-properties.js
```diff
@@ -140,11 +142,13 @@ const properties = {
...
  * @property {('left'|'center'|'right')=} [align='center'] - Horizontal alignment
  * @property {object=} style - Additional style options
  * @property {boolean=} [style.bold = true] - Bold text
  * @property {boolean=} [style.italic = false] - Italic text
  * @property {boolean=} [style.underline = false] - Underlined text
+ * @property {('responsive'|'relative'|'fixed')} [sizeBehavior='responsive'] - Setting to determine how the fontsize of the label is calculated.
```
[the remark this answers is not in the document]
niekvanstaveren: Yes I think so. Old objects would not have this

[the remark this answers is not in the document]
haxxmaxx: then there needs to be a `=` in there I think

src/ext.js
haxxmaxx: So what is the relative scale relative to? Seems like you are reusing the same slider, that goes from 0.2 to 1. it just seem very arbitary to me. I see that this is the simplest solution. Use the same slider, adjust the values so a "normal" label fits in a good way for those values

a-m-dev: Hey @haxxmaxx 👋 

I need to continue Niek's work on this one and i'm a bit confused with this comment of yours. what do you mean by this: "adjust the values so a "normal" label fits in a good way for those values"? 

The value is relative to **the box size** of the button, but do you mean that **the scale factor** here (numbers from 0.2 to 1) is seems a bit odd to you? 

are you concerned with the numbers that does not means something to the end user (I mean like if we stand in shoes of an end user (your concern would be) if it is relative -> then what 0.2 and 1 means? like relative to what? and what defines 0.2 and 1? )

am i understanding your question correctly? 🙂
if so, what would be the correct behaviour here and your suggestion, i would need some more info to fix it and get it going.

a-m-dev: we decided to get this one in, with out the Fluid options until we get more clarification on it, here is the extended work related to this PR: https://github.com/qlik-oss/sn-action-button/pull/331

Thanks @niekvanstaveren  for this 🙏

PR #327, opened 2022-11-17
feat: encode uri in open website action

mhs30:
1. Removed a regex that was not working properly, since the check could lead to mistakes, added a protocol check instead.
2. Added url encode to website navigation

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,6 @@
 
 Action button supernova for nebula.js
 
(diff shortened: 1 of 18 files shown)


mhs30:
After conversation with security-office:
1. it is expected to provide encoded url (url & path & parameters encoded) to [window.open](https://developer.mozilla.org/en-US/docs/Web/API/Window/open)  (not implemented by sn-action-button, will be implemented in this PR)
2. It is expected by customers and is security&privacy by default to break the back link between parent and child pages with noopener noreferrer  (not implemented sn-action-button, will not be implemented in this PR);

Caele (approved):

PR #332, opened 2022-11-21
feat: multi user feature

rileymd88:
- Added the ability for multiple users to trigger an automation
- Moved to using temporary bookmarks as they can be shared across users and have other useful features (TTL of 30 minutes, no need to publish/save like normal bookmarks)
- Added additional info to the body of the POST call to the automation


src/ext.js
```diff
@@ -155,19 +155,18 @@ export default function ext({ translator, shouldHide, senseNavigation }) {
                 },
                 // adds automation to actions and adds a dropdown property panel
                 // item to select the automation for the button to trigger
-                automation: {
+                automationId: {
                   type: 'string',
                   component: 'expression-with-dropdown',
                   translation: 'Object.ActionButton.Automation',
-                  ref: 'automation',
+                  ref: 'automationId',
```
Caele, line 162: changing the ref here will break all buttons already using this property. If the intention is to replace the old action then you need to make sure old objects that has `automation` saved are mapped correctly.

rileymd88: Hi @Caele 

I do handle the migration of this property to the new property here:
https://github.com/qlik-oss/sn-action-button/blob/31a1055794e9515bf87be1d2c22d94ad85f1b4a3/src/index.js#L70-L124

I also ensure if the property has not been migrated that it will still work here:
https://github.com/qlik-oss/sn-action-button/blob/31a1055794e9515bf87be1d2c22d94ad85f1b4a3/src/utils/actions.js#L304-L307

Caele: Still you'd want to hide the new functionality behind a feature flag, so want to keep the old code as well as the new?

rileymd88: Feature flag has now been added

src/object-properties.js
```diff
@@ -108,8 +108,12 @@ const properties = {
  * @property {string=} [value=''] - To select certain values in a field or set a variable. Required for types 'selectValues', 'selectMatchingValues', 'toggleSelect' and 'setVariable'
  * @property {string=} [variable=''] - Name of variable. Required for type 'setVariable'
  * @property {boolean=} [partial=false] - Set to true if you want to do a partial reload.
- * @property {string=} [automation=''] - ID of the automation. Required for type 'setVariable'
+ * @property {string=} [automationId=''] - ID of the automation. Required for type 'executeAutomation'
```
Caele, line 111: This will break api-governance, so might want to keep the old one.

src/utils/__tests__/actions.spec.js
```diff
@@ -306,34 +331,108 @@ describe('actions', () => {
...
+      jest.useFakeTimers('modern');
+      jest.setSystemTime(time);
+      const actionObject = actions.find((action) => action.value === 'executeAutomation');
+      await actionObject.getActionCall({ app, automation, automationId, automationTriggered, automationExecutionToken, automationPostData, buttonId })();
+      expect(global.fetch).toHaveBeenCalledTimes(4);
+      expect(global.fetch).toHaveBeenCalledWith(`../api/v1/users/me`);
+      expect(global.fetch).toHaveBeenCalledWith(`../api/v1/apps/${app.id}`);
+      expect(global.fetch).toHaveBeenCalledWith('../api/v1/csrf-token');
+      const automationData = {
+        guid: automationId,
```
lorenzvanherwaarden, line 388: **suggestion:** we can use `id: automationId` I think. The API still accepts `guid`, but it's an old property and `id` should be used if possible.

src/utils/actions.js
```diff
@@ -296,65 +298,119 @@ const actions = [
...
+              app: app.id,
+              bookmark: automationPostData ? bookmark : '',
+              sheet: await getSheetId(app, buttonId),
+              user: user.subject,
+              space: await getSpaceId(app.id),
+              tenant: user.tenantId,
+              time: new Date(),
+            }
+            const automationData = {
+              guid: automationId,
```
lorenzvanherwaarden, line 339: same comment here as above.

rileymd88: I changed this to id both here and in the tests

src/ext.js
```diff
@@ -155,33 +173,84 @@ export default function ext({ translator, shouldHide, senseNavigation }) {
...
                   type: 'boolean',
                   ref: 'automationPostData',
                   translation: 'Object.ActionButton.Automation.SendSelections',
                   show: (data) => checkShowAction(data, 'automation'),
                   defaultValue: false,
                 },
+                automationShowTriggered: {
+                  type: 'boolean',
+                  ref: 'automationShowTriggered',
+                  show: () => false,
```
cbt1, line 225: Why is it defined if it will be shown?

src/ext.js
```diff
@@ -1,9 +1,25 @@
...
 import { checkShowNavigation, getNavigationsList } from './utils/navigation-actions';
+import { getAutomation } from './utils/automationHelper';
 import propertyResolver from './utils/property-resolver';
 import importProperties from './utils/conversion';
 import luiIcons from './utils/lui-icons';
 
+let automationsList = null;
+
+const getAutomations = async () => {
+  if (!automationsList) {
```
cbt1, line 11: I'm guessing this is done so that `getAutomations` only fetches the list of automations once? Any particular reason why do that for `automationId` options but not for the `automation` options?

src/utils/actions.js
```diff
@@ -296,65 +298,119 @@ const actions = [
...
-            const itemInfo = await fetch(`../api/v1/items/${automation}`).then((response) => response.json());
-            const autoInfo = await fetch(`../api/v1/automations/${itemInfo.resourceId}`).then((response) =>
-              response.json()
-            );
-            let executePath = `../api/v1/automations/${autoInfo.id}/actions/execute?X-Execution-Token=${autoInfo.executionToken}`;
+      ({ app, automation, automationId, automationTriggered, automationExecutionToken, automationPostData, buttonId, multiUserAutomation }) =>
+        async () => {
+          if (multiUserAutomation) {
+            let automationUrl
+            if (typeof automation !== 'undefined' && automationId.length < 1) {
```
cbt1, line 305: Should be safe to do `automation !== undefined`

src/ext.js
rileymd88: This property is used to determine if we should show the automationTriggered option or not. The background is that the automationTriggered property should only be visible if the automation run mode === triggered. At the moment I set the automationShowTriggered property every time the user selects a new automationId - I did it this way to limit the amount of calls we need to make to the automation api.

src/ext.js
rileymd88: That is correct - I did not do this for the `automation` option as I just kept the old logic as it is (it will be no longer be used if the feature flag `SENSECLIENT_IM_1855_AUTOMATIONS_MULTI_USER` is present)

cbt1: Ok and the fetched data is expected to not became stale during its lifetime?

rileymd88: That is correct

src/utils/actions.js
rileymd88: I changed this in the latest commit

src/ext.js
cbt1: But you need to have it defined in the property panel? Can you not just set the value `automationShowTriggered` as you already do

src/utils/actions.js
```diff
@@ -296,65 +298,119 @@ const actions = [
...
-      async () => {
-        if (automation !== undefined) {
-          try {
-            automation = encodeURIComponent(automation);
-            const itemInfo = await fetch(`../api/v1/items/${automation}`).then((response) => response.json());
-            const autoInfo = await fetch(`../api/v1/automations/${itemInfo.resourceId}`).then((response) =>
-              response.json()
-            );
-            let executePath = `../api/v1/automations/${autoInfo.id}/actions/execute?X-Execution-Token=${autoInfo.executionToken}`;
+      ({ app, automation, automationId, automationTriggered, automationExecutionToken, automationPostData, buttonId, multiUserAutomation }) =>
```
haxxmaxx, line 301: please change the comment above to reflect the new arguments

src/ext.js
```diff
@@ -155,33 +173,84 @@ export default function ext({ translator, shouldHide, senseNavigation }) {
...
+                      data.automationExecutionToken = '';
+                    }
+                  },
+                },
+                automationTriggeredText: {
+                  label: `Object.ActionButton.Automation.RunModeTriggeredHelp`,
+                  component: 'text',
+                  show: (data) => checkShowAction(data, 'automation') && multiUserAutomation && data.automationShowTriggered,
+                },
+                automationExecutionToken: {
```
haxxmaxx, line 249: same here, should not be in this file if show is always false

src/object-properties.js
```diff
@@ -108,8 +108,13 @@ const properties = {
  * @property {string=} [value=''] - To select certain values in a field or set a variable. Required for types 'selectValues', 'selectMatchingValues', 'toggleSelect' and 'setVariable'
  * @property {string=} [variable=''] - Name of variable. Required for type 'setVariable'
  * @property {boolean=} [partial=false] - Set to true if you want to do a partial reload.
- * @property {string=} [automation=''] - ID of the automation. Required for type 'setVariable'
+ * @property {string=} [automation=''] - Item ID of the automation. Required for type 'executeAutomation'
+ * @property {string=} [automationId=''] - ID of the automation. Required for type 'executeAutomation'
```
haxxmaxx, line 112: I don't understand this. Are these the same thing? Hard to determine from the description. If not, are both of them are required for executeAutomation?

In general, I would strongly advice against renaming existing properties. I know from experience that handling different versions of the properties can be really messy when customers start migrating between sense versions etc.

haxxmaxx (commented):
I haven't reviewed everything, we need to clear the confusion about the property renaming first. Also, seems like your auto formatting is not working. Are you using vs code and have the prettier extension installed?

src/ext.js
rileymd88: I was under the impression that I needed to have it in the props panel in order to set it but I see that is not the case :) - I will remove it

src/utils/automationHelper.js
```diff
@@ -0,0 +1,129 @@
+export const getUser = async () => {
```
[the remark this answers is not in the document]
rileymd88: I will change this so snake case. I am using VS Code but was not using the prettier extension until now.

src/object-properties.js
rileymd88: I will write a better description in this file. 

The automation property which existed before was the item id coming from `api/v1/items?resourceType=automations` - saving this id in the properties is not the best thing to do as we need the automation id when triggering the automation and we were calling the API after every button click to get this information. I chose to leave the old automation prop as is and to create a new automationId prop which is using the actual automation id.

src/utils/actions.js
rileymd88: done in the latest commit

src/ext.js
rileymd88: done in the latest commit

src/object-properties.js
rileymd88: I added better descriptions in the latest commit

src/utils/automationHelper.js
```diff
@@ -0,0 +1,129 @@
...
+        return inputs.value === 'triggered';
+    }
+    return false;
+}
+
+
+export const checkIfTriggered = async (automation) => {
+    const startBlock = automation.workspace.blocks.find(block => block.type === 'StartBlock');
+    const inputs = startBlock.inputs.find(input => input.id === 'run_mode');
+    return inputs.value === 'triggered';
```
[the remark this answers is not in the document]
rileymd88: I actually do not use this function at all anymore so I removed it completely in the latest commit

src/utils/automationHelper.js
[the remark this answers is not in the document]
rileymd88: done now in the latest commit

src/object-properties.js
```diff
@@ -108,8 +108,13 @@ const properties = {
  * @property {string=} [value=''] - To select certain values in a field or set a variable. Required for types 'selectValues', 'selectMatchingValues', 'toggleSelect' and 'setVariable'
  * @property {string=} [variable=''] - Name of variable. Required for type 'setVariable'
  * @property {boolean=} [partial=false] - Set to true if you want to do a partial reload.
- * @property {string=} [automation=''] - ID of the automation. Required for type 'setVariable'
+ * @property {string=} [automation=''] - Item ID of the automation (the id field returned from /api/v1/items?resourceType=automations). Required for type 'executeAutomation' when feature flag SENSECLIENT_IM_1855_AUTOMATIONS_MULTI_USER is not enabled.
```
cbt1, line 111: I don't think we should mention feature flags in our API spec? @Caele

cbt1:
Not sure how this repo is setup when it comes do releasing but since you have changed the API spec, I'd advice you to look into generating a new spec file with those changes and commit to this PR.

src/ext.js
cbt1: Have you not pushed that commit?

rileymd88: sorry I missed this and thought it was there - should be there now in the latest commit

cbt1: Which commit?

rileymd88: this one here: https://github.com/qlik-oss/sn-action-button/pull/332/commits/508467210f40b98108faf14005658f8a77990e70

cbt1: Should not `automationExecutionToken` also be removed as it's never shown?

rileymd88:
> Not sure how this repo is setup when it comes do releasing but since you have changed the API spec, I'd advice you to look into generating a new spec file with those changes and commit to this PR.

Hi @cbt1 - I am not 100% sure what you mean by this - can you elaborate? Thanks!

cbt1:
> > Not sure how this repo is setup when it comes do releasing but since you have changed the API spec, I'd advice you to look into generating a new spec file with those changes and commit to this PR.
> 
> Hi @cbt1 - I am not 100% sure what you mean by this - can you elaborate? Thanks!

The API spec is generated from the JSDoc properties and you added one in this PR. But I just check, the release script takes care of making sure the API spec is generated. So you don't need to do anything!

cbt1:
@rileymd88 Can you fix the merge conflict? I have no other comments other than I think the flags should not be mention in the documentation. So I more or less ready to approve it.

rileymd88:
> @rileymd88 Can you fix the merge conflict? I have no other comments other than I think the flags should not be mention in the documentation. So I more or less ready to approve it.

@cbt1 I fixed the merge conflict and removed the flag mention from the docs 👍

cbt1 (approved):
LGTM but the action-button nor automation is not my area

PR #333, opened 2022-11-23
feat: enabled new styling panel for action button

rajani-varpe:
A new styling panel has been designed to be created in react, replacing the old style editor. This PR is about adding new styling panel(SENSECLIENT_IM_1525_BUTTON) through property definition and background options (BUTTON_BKG_OPTIONS+SENSECLIENT_IM_323_BACKGROUND_OPTIONS)

### DIFF
--- a/src/ext.js
+++ b/src/ext.js
@@ -30,8 +30,8 @@ const toggleOptions = [
 
 
 export default function ext({ translator, shouldHide, senseNavigation }) {
-  const stylingPanelEnabled = shouldHide?.isEnabled('SENSECLIENT_IM_1525_BUTTON');
-  const bkgOptionsEnabled = shouldHide?.isEnabled('BUTTON_BKG_OPTIONS');
+  const stylingPanelEnabled = shouldHide.isEnabled && shouldHide.isEnabled('SENSECLIENT_IM_1525_BUTTON');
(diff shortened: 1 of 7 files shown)


Caele (approved):

src/ext.js
```diff
@@ -26,7 +27,11 @@ const toggleOptions = [
   },
 ];
 
+
+
 export default function ext({ translator, shouldHide, senseNavigation }) {
+  const stylingPanelEnabled = shouldHide.isEnabled && shouldHide.isEnabled('SENSECLIENT_IM_1525_BUTTON');
+  const bkgOptionsEnabled = shouldHide.isEnabled && shouldHide.isEnabled('BUTTON_BKG_OPTIONS');
```
haxxmaxx, line 34: Is this feature flag new? Does it belong to the same improvement? In that case, add the `IM_1525` to the flag name

rajani-varpe: This feature flag is used for the same purpose like we did back in KPI: to enable/disable the background options on styling panel for this chart and the KPI uses the flag `KPI_BKG_COLOR`, so we thought of maintaining the same name consistency only renaming the `COLOR` part to `OPTIONS`, also the background options belong to another IM i.e IM-323.

haxxmaxx (approved):

PR #339, opened 2022-12-22
feat: enabled chart tab for button object along with maintaining backward compatibility of properties

rajani-varpe:
This PR intends to enable the chart tab in button object along with backward compatibility of editor properties except the background Image using Media Library.

Background Image using Media Library is working fine in Styling Panel, the action item to get this working with backward compatibility is to add the correct condition in MediaLibrary component. This functionality works fine with "**args.properties**" (MediaLibrary Component)but later this has been changed to "**args.handler.layout.component**" which tend to fail as we do not have any "components" in style editor. 

So this condition needs to be handled to get this working properly.

### 1. Styling Panel

image

### 2. Style Editor - without FF

image

### 3. Styling Panel - with FF

image


Donya-qlik:
@AmeliaAndersson @LiKang6688 @niekvanstaveren Could you please start reviewing this PR? I'll take care of this PR since the responsible person is not working with Qlik anymore. Thanks!

**Note**: Rajani has mentioned in his explanations that Media Library is not working with `args.layout.properties`. I'm not sure I understood what he meant. Media Library works properly in sense-client as an extension as far as I tested it. Let me know if you have any idea about it. Thanks.

src/styling-panel-definition.js
```diff
@@ -11,7 +44,406 @@ const getStylingPanelDefinition = (bkgOptionsEnabled) => ({
       ref: 'components',
       useGeneral: true,
       useBackground: bkgOptionsEnabled,
+      items: {
+        labelSection: {
+          translation: 'Label',
+          component: 'panel-section',
+          show(data, handler, args) {
+            console.log("show=>labelSection=>", args)
```
niekvanstaveren, line 52: Remove all console.log statements before merging

src/styling-panel-definition.js
```diff
@@ -11,7 +44,406 @@ const getStylingPanelDefinition = (bkgOptionsEnabled) => ({
       ref: 'components',
       useGeneral: true,
       useBackground: bkgOptionsEnabled,
+      items: {
+        labelSection: {
+          translation: 'Label',
```
niekvanstaveren, line 49: Is this a real translation?

src/styling-panel-definition.js
```diff
@@ -11,7 +44,406 @@ const getStylingPanelDefinition = (bkgOptionsEnabled) => ({
...
+          component: 'panel-section',
+          show(data, handler, args) {
```
niekvanstaveren, line 51: What is this show function supposed to do? It just return true in the end, but there is a lot of logic in here which should not be here

src/styling-panel-definition.js
```diff
@@ -11,7 +44,406 @@ const getStylingPanelDefinition = (bkgOptionsEnabled) => ({
...
+              obj.label = args?.handler?.layout?.style?.label;
+              args.properties.components.push(obj)
+            }
+            return true;
+          },
+          items: {
+            labelItem: {
+              component: 'items',
+              ref: 'components',
+              key: 'actionbutton',
```
niekvanstaveren, line 102: I don't think these need to be in the components array. Only when this is something that could be generic it should be placed in components. The key `actionbutton` already shows that it is probably not generic

src/styling-panel-definition.js
```diff
@@ -11,7 +44,406 @@ const getStylingPanelDefinition = (bkgOptionsEnabled) => ({
...
+                      translation: 'Common.Expression',
+                      show: (data) => propertyResolver.getValue(data, 'font.useColorExpression'),
+                    },
+                  }
+                }
+              },
+            },
+          },
+        },
+        bgColorSection: {
```
niekvanstaveren, line 178: Why do we have this now both in the general tab and the chart tab?

niekvanstaveren (changes requested):
A couple things here really need to be fixed. I would actually suggest starting a new PR for this.

Have not tested things myself, but I am pretty sure that there will be some issues when using the new options. For instance what would happen if I have a background selected in the old setup and now want to change the font color with the feature flag active. Do I still get the old background settings applied?

PR #346, opened 2023-02-08
feat: enable Chart styling panel for label styling

Donya-qlik:
**In this PR:** 
Enable styling panel chart tab which includes the listed items: (the backward compatibility has also been considered)
- Enabling Chart tab
- FontFamily dropdown
- Font style
- Text align
- FontSize
- FontColor (single and expression)
- Reset the Label section option

**Feature-flag:** SENSECLIENT_IM_1525_STYLINGPANEL_BUTTON
**Note:** To be able to test it locally, sense-client needs to be linked to analytic changes on my branch [here](https://github.com/qlik-trial/analytics/pull/401) and for getting new translations to [sense-client](https://github.com/qlik-trial/sense-client/pull/23139)

**Unsolved Issues:**

- The font size tooltip looks like moving behind the font styling wrapper in its current location. If we consider LayoutBehavior component which is supposed to be set between these two, it will not behave the same so it might not be the matter of issue for now. LayoutBehavior will be added in a separate PR.

- ResetAll which resets all the components on the panel at the same time, doesn't work. This function works with component key that has not been considered in this feature. I'll consider it in another PR.

https://user-images.githubusercontent.com/20701546/217811775-404c71a3-0d92-498c-ac49-0faf6d4d831f.mp4

### DIFF
--- a/src/object-properties.js
+++ b/src/object-properties.js
@@ -63,6 +63,7 @@ const properties = {
   style: {
     label: DEFAULTS.LABEL,
     font: {
+      fontFamily: DEFAULTS.FONT_FAMILY,
       useColorExpression: false,
       color: DEFAULTS.COLOR,
       colorExpression: '',
@@ -151,6 +152,7 @@ const properties = {
 
 /**
  * @typedef {object} Font
+ * @property {string=} [fontFamily='Source Sans Pro'] - Font Family of the label
  * @property {boolean=} [useColorExpression=false] - Set to true to use color expression
  * @property {PaletteColor=} color - Color defined by index or hex code, needed if useColorExpression is false
  * @property {ColorExpression=} colorExpression - Color defined by expression, needed if useColorExpression is true
--- a/src/styling-panel-definition.js
+++ b/src/styling-panel-definition.js
@@ -1,4 +1,86 @@
-const getStylingPanelDefinition = (bkgOptionsEnabled) => ({
+import styleDefaults from './style-defaults';
+import propertyResolver from './utils/property-resolver';
+import { fontFamilyOptions, colorOptions } from './utils/style-utils';
+
+export const getStyleEditorDefinition = () => ({
+  items: {
+    labelSection: {
+      component: 'panel-section',
+      translation: 'Common.Label',
+      items: {
+        labelItem: {
+          items: {
+            fontFamily: {
+              component: 'dropdown',
+              ref: 'style.font.fontFamily',
+              options: fontFamilyOptions,
+              defaultValue: styleDefaults.FONT_FAMILY,
+            },
+            fontWrapper: {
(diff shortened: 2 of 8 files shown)


Donya-qlik:
There are still couple of issues: 

- The font size tooltip looks like moving behind the font styling wrapper in its current location. If we consider `LayoutBehavior` component which is supposed to be set between font styling wrapper and font size, it will not behave the same so it might not be a matter for this PR. `LayoutBehavior` will be added in the later PRs. 
- ResetAll which resets all the components at the same time, doesn't work. This function works with component key that has not been considered in this feature. Without keys it might need some nested for-loops to reach to each component and reset them one by one which doesn't sound practical so it should be discussed.

src/styling-panel-definition.js
```diff
@@ -1,4 +1,86 @@
...
+              component: 'inline-wrapper',
+              type: 'items',
+              grouped: true,
+              items: {
+                fontStyle: {
+                  component: 'font-style-buttons',
+                  width: false,
+                  ref: 'style.font.style',
+                  defaultValue: ['bold'],
+                  matchValueFormat: true,
```
Donya-qlik, line 29: I might change this to `formatValue` instead.
In the new styling button component (font-style-buttons), the value is an object when in the old version it is in an array. Using these in different places requires a converter that converts the values to the right format. Whenever style button group is used this property `matchValueFormat` should be set to true.

niekvanstaveren: Actually I think we should should use the existing `type` option.
So add `type: 'array'` here instead

src/object-properties.js
```diff
@@ -63,6 +63,7 @@ const properties = {
   style: {
     label: DEFAULTS.LABEL,
     font: {
+      fontFamily: DEFAULTS.FONT_FAMILY,
```
niekvanstaveren, line 66: Do we really need to set this one? Would be great if we can without so the theme could override this

src/styling-panel-definition.js
```diff
@@ -1,4 +1,86 @@
...
+    labelSection: {
+      component: 'panel-section',
+      translation: 'Common.Label',
+      items: {
+        labelItem: {
+          items: {
+            fontFamily: {
+              component: 'dropdown',
+              ref: 'style.font.fontFamily',
+              options: fontFamilyOptions,
```
niekvanstaveren, line 16: Would be great if we can have this in the client/analytics instead. Otherwise we need to keep a copy of the array in every chart that we want to use it. For now this is fine

niekvanstaveren (approved):
Yes, this looks great!

src/object-properties.js
Donya-qlik: How can I give reference to it then? font family is a dropdown with an option of different fonts

PR #347, opened 2023-02-09
feat: add background options and borders to styling panel

Donya-qlik:
The background and border styling options have been added to the action-button chart styling panel, it includes the listed items.  

- Background color
- Background Image using MediaLibrary
- Border
- Reset the section

Feature-flag: SENSECLIENT_IM_1525_STYLINGPANEL_BUTTON
Note: To be able to test it locally, sense-client needs to be linked to analytic changes on my branch https://github.com/qlik-trial/analytics/pull/401 and for getting new translations to https://github.com/qlik-trial/sense-client/pull/23139

Unsolved Issues of the current PR:

- Color Picker preview boxes are not showing the default colors that should be fixed in the coming PRs.
- The position grid does not show focus on center at the beginning. It is because the old and new version of positionGrid component take different values for the default position. The current value is `centerCenter` whereas it is 'center-center' with the new component. 

https://user-images.githubusercontent.com/20701546/219906767-1bd6c05c-8334-4962-8fe2-adcfafdacf49.mp4

### DIFF
--- a/src/object-properties.js
+++ b/src/object-properties.js
@@ -65,7 +65,7 @@ const properties = {
     font: {
       fontFamily: DEFAULTS.FONT_FAMILY,
       useColorExpression: false,
-      color: DEFAULTS.COLOR,
+      color: DEFAULTS.FONT_COLOR,
       colorExpression: '',
       size: DEFAULTS.FONT_SIZE,
       style: DEFAULTS.FONT_STYLE,
@@ -75,6 +75,7 @@ const properties = {
       useColorExpression: false,
       color: DEFAULTS.COLOR,
       colorExpression: '',
+      mode: DEFAULTS.BGIMAGE_MODE,
       useImage: false,
       url: {
         qStaticContentUrlDef: {
@@ -169,11 +170,12 @@ const properties = {
  * @property {boolean=} [useColorExpression=false] - Set to true to use color expression
  * @property {PaletteColor=} color - Color defined by index or hex code, needed if useColorExpression is false
  * @property {ColorExpression=} colorExpression - Color defined by expression, needed if useColorExpression is true
+ * @property {('none'|'media')=} [mode = 'none'] - Toggles the background image, set to media to show the image
  * @property {boolean=} [useImage=false] - Set to true to show background image
  * @property {object=} url - Contains the URL for the background image
  * @property {object=} url.qStaticContentUrlDef
  * @property {string=} url.qStaticContentUrlDef.qUrl - URL represented as a string
- * @property {('topLeft'|'centerLeft'|'bottomLeft'|'topCenter'|'centerCenter'|'bottomCenter'|'topRight'|'centerRight'|'bottomRight')} [position='centerCenter'] - Image position
+ * @property {('topLeft'|'centerLeft'|'bottomLeft'|'topCenter'|'centerCenter'|'bottomCenter'|'topRight'|'centerRight'|'bottomRight'|'top-left'|'center-left'|'bottom-left'|'top-center'|'center-center'|'top-right'|'center-right'|'bottom-right')} [position='center-center'] - Image position
  * @property {('auto'|'alwaysFit'|'fitWidth'|'fitWidth'|'fitHeight'|'fill'|'alwaysFill')=} [size='auto'] - Size of the image, relative to the button
  */
 
--- a/src/styling-panel-definition.js
+++ b/src/styling-panel-definition.js
@@ -1,6 +1,6 @@
 import styleDefaults from './style-defaults';
 import propertyResolver from './utils/property-resolver';
-import { fontFamilyOptions, colorOptions } from './utils/style-utils';
+import { fontFamilyOptions, colorOptions, toggleOptions } from './utils/style-utils';
 
 export const getStyleEditorDefinition = () => ({
   items: {
@@ -56,10 +56,12 @@ export const getStyleEditorDefinition = () => ({
                   options: colorOptions,
                 },
                 colorPicker: {
-                  component: 'color-picker',
                   type: 'object',
+                  component: 'color-picker',
                   ref: 'style.font.color',
                   translation: 'properties.color',
+                  defaultValue: styleDefaults.FONT_COLOR,
+                  disabledNone: true,
                   width: 12,
                   show: (data) => !propertyResolver.getValue(data, 'style.font.useColorExpression'),
                 },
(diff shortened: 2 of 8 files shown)


src/utils/style-utils.js
```diff
@@ -48,3 +81,19 @@ export const fontFamilyOptions = FONT_FAMILIES.map((font) => ({
   value: font,
   label: getFirstFont(font),
 }));
+
+export const getColor = (
```
haxxmaxx, line 85: since there is a color-utils, doesn't this fit better there?

src/utils/style-formatter.js
```diff
@@ -2,71 +2,39 @@ import colorUtils from './color-utils';
...
     const fontStyle = font.style || DEFAULTS.FONT_STYLE;
     fontStyle.bold && (styles += formatProperty('font-weight', 'bold'));
     fontStyle.italic && (styles += formatProperty('font-style', 'italic'));
     // background
+    const primaryColor = theme.getDataColorSpecials().primary;
+    const backgroundColor = getColor(background, primaryColor, theme);
     styles += formatProperty('background-color', backgroundColor);
-    if (background.useImage && background.url.qStaticContentUrl) {
+    // backgroundImage
+    const {useImage} = background;
```
haxxmaxx, line 26: Are you using the linter? It should do this automatically
```suggestion
    const { useImage } = background;
```

src/object-properties.js
```diff
@@ -159,12 +160,13 @@ const properties = {
  * @property {boolean=} [useColorExpression=false] - Set to true to use color expression
  * @property {PaletteColor=} color - Color defined by index or hex code, needed if useColorExpression is false
  * @property {ColorExpression=} colorExpression - Color defined by expression, needed if useColorExpression is true
+ * @property {string=} [mode = 'none'] - Set to media library to show background image
  * @property {boolean=} [useImage=false] - Set to true to show background image
```
haxxmaxx, line 164: wait, do we need both here? looks like useImage and mode does the same thing

haxxmaxx (approved):

haxxmaxx:
woops didn't mean to approve, seems like I can't revert it. Will approve again when comments are taken care of

src/utils/style-utils.js
Donya-qlik: I found color-utils it more related to set css colors and color code stuff... but sure I can move it over there.

src/object-properties.js
Donya-qlik: Well...the old component of use-background-image in the main panel(not styling one) is a switch that here it is defined by a boolean useImage. The new one is a dropdown that is called mode, and has two options to show/hide the image, so should keep both.

src/utils/style-utils.js
```diff
@@ -21,6 +24,36 @@ const FONT_FAMILIES = [
...
+export const backgroundPosition = {
+  'top-left': 'top left',
+  'center-left': 'center left',
+  'bottom-left': 'bottom left',
+  'top-center': 'top center',
+  'center-center': 'center center',
+  'bottom-center': 'bottom center',
+  'top-right': 'top right',
+  'center-right': 'center right',
+  'bottom-right': 'bottom right',
```
Donya-qlik, line 42: @niekvanstaveren I tried with replacing the these values with percentage ones (e.x: 50% 50%) but it didn't work, clicking around position component didn't change the image position so I kept it as it is.

src/styling-panel-definition.js
```diff
@@ -52,14 +52,17 @@ export const getStyleEditorDefinition = () => ({
                   type: 'boolean',
                   translation: 'properties.fontColor',
                   component: 'dropdown',
+                  defaultValue: false,
```
niekvanstaveren, line 55: I think we made a mistake here in the past. A dropdown and boolean should not be together

src/styling-panel-definition.js
```diff
@@ -77,6 +80,191 @@ export const getStyleEditorDefinition = () => ({
...
+                  value: 'media',
+                  translation: 'MediaLibrary.Header',
+                },
+              ],
+              change(data) {
+                const bgImageComp = data.style.background;
+                bgImageComp
+                  ? (bgImageComp.url = { qStaticContentUrlDef: '' })
+                  : (data.style.background = { url: { qStaticContentUrlDef: '' } });
+              },
```
niekvanstaveren, line 139: Why resetting the background image here?

src/styling-panel-definition.js
```diff
@@ -77,6 +80,191 @@ export const getStyleEditorDefinition = () => ({
...
+                  translation: 'properties.backgroundImage.sizeStretch',
+                },
+                {
+                  value: 'alwaysFill',
+                  translation: 'properties.backgroundImage.sizeAlwaysFill',
+                },
+              ],
+              change: (data) => {
+                let currentPosition = propertyResolver.getValue(data, 'style.background.position');
+                if (currentPosition) currentPosition = styleDefaults.BGIMAGE_POSITION;
```
niekvanstaveren, line 179: Do we need to reset the position every time we change the size?

src/styling-panel-definition.js
```diff
@@ -77,6 +80,191 @@ export const getStyleEditorDefinition = () => ({
...
+      translation: 'properties.border',
+      items: {
+        backgroundBorderItem: {
+          items: {
+            useBorder: {
+              type: 'boolean',
+              component: 'switch',
+              ref: 'style.border.useBorder',
+              translation: 'properties.border.use',
+              defaultValue: toggleOptions[1].value,
```
niekvanstaveren, line 211: Just `false` would be easier I think

niekvanstaveren (approved):
Some small comments. Please have a look at them

src/utils/__tests__/style-formatter.spec.js
```diff
@@ -128,7 +192,7 @@ describe('style-formatter', () => {
           },
         },
       };
-      const formattedStyle = styleFormatter.getStyles({ style, disabled, theme, app });
+      let formattedStyle = styleFormatter.getStyles({ style, disabled, theme, app });
```
niekvanstaveren, line 195: This should be a const (the lint is also complaining about it)

src/styling-panel-definition.js
Donya-qlik: Ah I see I forgot to remove it from here, I did in the later PRs. I'll fix it here.

src/styling-panel-definition.js
Donya-qlik: Changing happens when you select another size mode and each option enables/disables some positions spots. Thus, yes it puts it back to center and let you choose the new position.

src/styling-panel-definition.js
Donya-qlik: Oh weird, I had fixed some details before but they are still here... I'll fix it.

PR #353, opened 2023-02-27
refactor: format fixes

niekvanstaveren:


cbt1 (approved):
LGTM but the commit type of this PR should be `refactor` or `chore` and  not `fix`. I'd lean more towards `chore`.  Given no actual bug have been fixed?

niekvanstaveren:
> LGTM but the commit type of this PR should be `refactor` or `chore` and not `fix`. I'd lean more towards `chore`. Given no actual bug have been fixed?

In that case I think refactor. chore would be more for build/dev processes.

PR #355, opened 2023-03-01
fix: changing mode from media to none deletes the url value

Donya-qlik:
This PR fixed the issue of 
- Changing the mode from 'media' to 'none', the background image URL will be deleted. However, the URL should only be deleted if the panel section is reset.

### DIFF
--- a/src/utils/__tests__/style-formatter.spec.js
+++ b/src/utils/__tests__/style-formatter.spec.js
@@ -1,5 +1,6 @@
 import styleFormatter from '../style-formatter';
 import defaultValues from '../../__tests__/default-button-props';
+import { getStyleEditorDefinition } from '../../styling-panel-definition';
 
 describe('style-formatter', () => {
   describe('getStyles', () => {
@@ -24,7 +25,7 @@ describe('style-formatter', () => {
       const d = defaultValues();
       theme = d.theme;
       layout = d.layout;
-      style = JSON.parse(JSON.stringify(layout.style));
+      style = layout.style;
       element = {
         offsetHeight: 200,
         offsetWidth: 100,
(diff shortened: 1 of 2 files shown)


src/utils/__tests__/style-formatter.spec.js
```diff
@@ -129,6 +129,18 @@ describe('style-formatter', () => {
...
+      style.background.mode = 'media';
+      style.background.url.qStaticContentUrl = { qUrl: someUrl };
+      style.background.mode = 'none';
+      const formattedStyle = styleFormatter.getStyles({ style, disabled, theme, app });
+      expect(formattedStyle.includes('background-image:')).toBe(false);
+      style.background.mode = 'media';
+      styleFormatter.getStyles({ style, disabled, theme, app });
+      expect(style.background.url.qStaticContentUrl.qUrl).toEqual(someUrl);
+    });
+
```
eliseeborn, line 143: does this test really test what you expect it to? if you revert your changes in styling-panel-definition.js does it fail?
I'm just asking because I'm not sure the change-function is called when the values aren't changed through the styling panel.

Donya-qlik: True! I'll find another way to test it.

Donya-qlik: @eliseeborn I spent some time to find the right scenario with change-function but then I realized that I had totally missed the point that there is no reset on this component individually and resetting the panel section does nothing to do with change-function, even deleting the image(removing it by clicking the cross sign) doesn't call it. It is called whenever the user selects a new option from the dropdown that corresponds to 'style.background.mode', and it updates the component's internal state based on the new data. Thus, it doesn't matter if we switch the conditions, it takes the one that fits to the condition and it doesn't change the data state. Sorry if I made it confusing by using 'reset' in my explanations. I'll modify the commit message and the description.

eliseeborn: this test still doesn't work though, changing the mode from inside the code won't call the change-function either

Donya-qlik: Yes the test is silly.... 
I modified it to a better scenario. However, it's not really possible to render the component and check the real behavior. I just called the change function at each new state and verified the results backgroundImage availability. Does it make more sense ?

eliseeborn:
I think you just need to merge master to fix the api compliance check

niekvanstaveren:
> I think you just need to merge master to fix the api compliance check

Probably not. We need to get an override for api governance to work again

PR #360, opened 2023-03-06
chore: set unit test threshold

niekvanstaveren:


cbt1:
Looks like this PR is also changing the coverage :oops:, so you might want to update the config file to either include only relevant src files or at least exclude the jest config file

cbt1:
I wonder where coveralls gets that new coverage number from? Look at the coverage reported generated during the build does not produce any that would result in that number from what I can tell

niekvanstaveren:
> I wonder where coveralls gets that new coverage number from? Look at the coverage reported generated during the build does not produce any that would result in that number from what I can tell

No me neither. I'm gonna play a bit locally with this before merging. See if I can get this correct

niekvanstaveren:
> I wonder where coveralls gets that new coverage number from? Look at the coverage reported generated during the build does not produce any that would result in that number from what I can tell

It ended up adding some files to the coverage I think that were not included before. Added some to the excluded path that were only containing constants

cbt1 (approved):

PR #362, opened 2023-03-08
chore: format check in ci

niekvanstaveren:


cbt1:
I don't think we run prettier check on CI on most of our other projects. Why would we need it here?

niekvanstaveren:
> I don't think we run prettier check on CI on most of our other projects. Why would we need it here?

I would say we should probably have it in the other projects as well. Have a PR up for the org chart now as well. Especially in open source projects I think we should try to enforce this (when people do not have prettier setup properly).

cbt1:
> 

I think we just use husky to run estlint before commiting to catch those issues. As this project also has eslint prettier plugin that would work here too.

cbt1 (approved):
IMO I would go with a pre-commit check instead but if you prefer this solution I'm fine with that

niekvanstaveren:
> IMO I would go with a pre-commit check instead but if you prefer this solution I'm fine with that

Ah right. Yes that sounds better indeed. Thanks 😄

PR #364, opened 2023-03-14
fix: remove automationshowtriggered

niekvanstaveren:
This will probably break the API check. We should not have exposed this in the first place. Might have to do a major button release because of this


cbt1:
Has there been a release with that property?

niekvanstaveren:
> Has there been a release with that property?

Yes, sadly there was

cbt1 (approved):

PR #365, opened 2023-03-16
fix: get base url from websocket connection url

niekvanstaveren:
Think I got all needed fetch calls here. Will have to do some testing around this, but wanted to get the PR open for feedback


rileymd88 (approved):

niekvanstaveren:
This will not work without authentication

PR #369, opened 2023-03-29
feat: add font size behavior dropdown

Donya-qlik:
This PR is mostly based on @niekvanstaveren and @a-m-dev developments of font size behavior. [#312](https://github.com/qlik-oss/sn-action-button/pull/312), [#331](https://github.com/qlik-oss/sn-action-button/pull/331)
The font size behavior drop down is added which provides three options of `responsive`, `fluid `and `fixed `font scaling.

**Options definitions:**
- **Fixed** in which the font size is independent of the box size and the length of the text
- **Fluid** in which the font size adapts to the size of the box
- **Responsive** in which the font size adapts to the size of the box and the length of text.

### DIFF
--- a/src/utils/style-formatter.js
+++ b/src/utils/style-formatter.js
@@ -2,7 +2,7 @@ import colorUtils from './color-utils';
 import luiIcons from './lui-icons';
 import { getImageUrl } from './url-utils';
 import DEFAULTS from '../style-defaults';
-import { backgroundSize, backgroundPosition } from './style-utils';
+import { backgroundSize, backgroundPosition, setTextFontSize } from './style-utils';
 
 const formatProperty = (path, setting) => `${path}: ${setting};`;
 
@@ -56,13 +56,13 @@ export default {
     const { icon = {}, font = {}, label = DEFAULTS.LABEL } = style;
     // text element wrapping label and icon
     const text = document.createElement('text');
-    text.style.whiteSpace = 'nowrap';
+    text.style.whiteSpace = 'pre';
     text.style.fontFamily = style.font.fontFamily || theme.getStyle('', '', 'fontFamily') || DEFAULTS.FONT_FAMILY;
 
     // label
     const textSpan = document.createElement('span');
     textSpan.textContent = label;
-    textSpan.style.whiteSpace = 'nowrap';
+    textSpan.style.whiteSpace = 'pre';
     textSpan.style.textOverflow = 'ellipsis';
     textSpan.style.overflow = 'visible';
     font.style && font.style.underline && (textSpan.style.textDecoration = 'underline');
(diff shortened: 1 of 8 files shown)


a-m-dev:
I'm afraid to say that rendering tests results is questioning all of our hypothesis on responsive and fluid type. 

Also, I saw this in rendering fixture: 
```js
font: {
  size: 0.46,
  useColorExpression: false,
  color: {
    index: 8,
    color: '#8a85c6',
  },
  colorExpression: '',
  style: {
    bold: true,
    italic: true,
    underline: true,
  },
  align: 'center',
},
```
Seems like [according to this line in your changes](https://github.com/qlik-oss/sn-action-button/pull/369/files#diff-58f9ee076a4851f28c77d33d314c3fa7ab818ab4e442d4e8c061a0ddbde657e4R71 ) the default behaviour is **responsive** in rendering tests, but why text does not adapts with bigger button wrapper is the question and need to be figured out!

src/utils/style-formatter.js
```diff
@@ -80,33 +80,39 @@ export default {
...
-    if (font.style && font.style.italic) {
-      if (hasIcon) {
-        text.style.fontSize = `${Math.max(newFontsize * fontScale * 0.84, 8)}px`;
-        text.children[0].style.marginRight = `${text.offsetWidth * 0.04}px`;
-        text.children[1].style.marginRight = `${text.offsetWidth * 0.04}px`;
-      } else {
-        text.style.fontSize = `${Math.max(newFontsize * fontScale * 0.9, 8)}px`;
-        text.children[0].style.marginRight = `${text.offsetWidth * 0.02}px`;
+    // font size adapts to the size of the box and the length of text
+    if (Object.keys(font).length === 0 || font.sizeBehavior === 'responsive') {
```
a-m-dev, line 84: is this check possible? 
```js
Object.keys(font).length === 0
```
According to `src/object-properties.js` you are setting a default value for entire `font` right? which includes default value for `font.sizeBehavior` that makes it to be 'responsive' always (if i'm not mistaken)

src/utils/style-formatter.js
```diff
@@ -80,33 +80,39 @@ export default {
...
+      console.log(Object.keys(font).length);
```
a-m-dev, line 85: oops!

a-m-dev:
Also make sure to test this changes with `hasIcon` scenario as well, because that is our blind spot in integration tests!

src/utils/style-formatter.js
Donya-qlik: When `resetAll` from StylingPanel is called all font properties will be deleted and font item will be an empty object `font: {}`. We want to be sure it takes the first option which is the default one(responsive).

src/utils/style-formatter.js
Donya-qlik: oh yes, will remove it

Donya-qlik:
> Interesting, rendering tests are passing in my local (with your changes by running `test:local:rendering` command) -> snapshot comparison is green, but not sure why it fails in CI
> 
> Screenshot 2023-03-30 at 11 10 41

Mines are failed on CI, haven't rendered them locally

Donya-qlik:
> I'm afraid to say that rendering tests results is questioning all of our hypothesis on responsive and fluid type.
> 
> Also, I saw this in rendering fixture:
> 
> ```js
> font: {
>   size: 0.46,
>   useColorExpression: false,
>   color: {
>     index: 8,
>     color: '#8a85c6',
>   },
>   colorExpression: '',
>   style: {
>     bold: true,
>     italic: true,
>     underline: true,
>   },
>   align: 'center',
> },
> ```
> 
> Seems like [according to this line in your changes](https://github.com/qlik-oss/sn-action-button/pull/369/files#diff-58f9ee076a4851f28c77d33d314c3fa7ab818ab4e442d4e8c061a0ddbde657e4R71) the default behaviour is **responsive** in rendering tests, but why text does not adapts with bigger button wrapper is the question and need to be figured out!


> I'm afraid to say that rendering tests results is questioning all of our hypothesis on responsive and fluid type.
> 
> Also, I saw this in rendering fixture:
> 
> ```js
> font: {
>   size: 0.46,
>   useColorExpression: false,
>   color: {
>     index: 8,
>     color: '#8a85c6',
>   },
>   colorExpression: '',
>   style: {
>     bold: true,
>     italic: true,
>     underline: true,
>   },
>   align: 'center',
> },
> ```
> 
> Seems like [according to this line in your changes](https://github.com/qlik-oss/sn-action-button/pull/369/files#diff-58f9ee076a4851f28c77d33d314c3fa7ab818ab4e442d4e8c061a0ddbde657e4R71) the default behaviour is **responsive** in rendering tests, but why text does not adapts with bigger button wrapper is the question and need to be figured out!

Sorry couldn't follow you properly.... but I guess you didn't see the line I changed the `sizeBehavior `to fluid/fixed in new test cases. All responsive scenarios are working as they did (they are not added by me) and the new ones are working as I expected. The text will not adapt to the bigger button wrapper if it is set to fixed size. The size you see there 0.46 is the font size slider value that I later multiplied it by 100 to get the value in `px` unit. If it is not what you had questioned on then let me know which testcase you have issue with, there are few : )

src/utils/style-formatter.js
a-m-dev: ok then, this issue goes back to styling panel, that reset should not empty any of the layout property -> it should reset it back to where it was? (in case of default values like this one IMO || at lease it might has a kind rigid default values at the base level to some extent that does not hurts us with checks like this)

a-m-dev:
I was pointing to rendering tests in my comment, unit tests are green in CI

src/utils/__tests__/__snapshots__/style-utils.spec.js
```diff
@@ -0,0 +1,59 @@
...
+    setTextFontSize(text, font, textFontSize, hasIcon);
+    expect(text.style.fontSize).toBe('10.56px');
+    expect(text.children[0].style.marginRight).toBe('0.8px');
+  });
+
+  it('should set font size for non-italic font without icon', () => {
+    hasIcon = false;
+    font.style.italic = false;
+    setTextFontSize(text, font, textFontSize, hasIcon);
+    expect(text.style.fontSize).toBe('11.040000000000001px');
```
a-m-dev, line 57: This seems not correct, I would suggest incorporating `parseFloat(CALCULATED_FONT_SIZE).toFixed(2)` somewhere in `setTextFontSize` function.

a-m-dev:
also you've put `style-utils.psce.ts` file in your `__snapshot__` tests folders. pls move it up one level in folder structure to `__tests__` folder.

src/utils/__tests__/__snapshots__/style-utils.spec.js
a-m-dev: something like this: 
```ts
export const setTextFontSize = (text, font, textFontSize, hasIcon) => {
  let fontSize;
  if (font.style && font.style.italic) {
    if (hasIcon) {
      fontSize = Math.max(textFontSize * 0.84, 8);
      text.children[0].style.marginRight = `${text.offsetWidth * 0.04}px`;
      text.children[1].style.marginRight = `${text.offsetWidth * 0.04}px`;
    } else {
      fontSize = Math.max(textFontSize * 0.9, 8);
      text.children[0].style.marginRight = `${text.offsetWidth * 0.02}px`;
    }
  } else if (hasIcon) {
    fontSize = Math.max(textFontSize * 0.88, 8);
    text.children[0].style.marginRight = `${text.offsetWidth * 0.04}px`;
  } else {
    fontSize = Math.max(textFontSize * 0.92, 8);
  }
  
  text.style.fontSize = `${parseFloat(fontSize).toFixed(2)}px`;
};
```

haven't tried it out since i don't all project glued together, but make sure to give it a try first!

a-m-dev (approved):
LGTM

PR #374, opened 2023-04-11
fix(QB-18860): button fontSizeBehavior is not compatible

Donya-qlik:
**Issue:**
sizeBehavior doesn't have a compatibility with the old button

**Fix:**
Refactor the conditions to set any conditions other than `fixed` and `fluid` to `responsive ` which is the default value.

More rendering-tests added via this PR: #375

### DIFF
--- a/src/object-properties.js
+++ b/src/object-properties.js
@@ -159,7 +159,7 @@ const properties = {
  * @property {boolean=} [style.bold = true] - Bold text
  * @property {boolean=} [style.italic = false] - Italic text
  * @property {boolean=} [style.underline = false] - Underlined text
- * @property {('responsive'|'fluid'|'fixed')} [sizeBehavior='responsive'] - Setting to determine how the fontsize of the label is calculated.
+ * @property {('responsive'|'relative'|'fixed')} [sizeBehavior='responsive'] - Setting to determine how the fontsize of the label is calculated.
  */
 
 /**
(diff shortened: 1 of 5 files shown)


src/object-properties.js
```diff
@@ -159,7 +159,7 @@ const properties = {
  * @property {boolean=} [style.bold = true] - Bold text
  * @property {boolean=} [style.italic = false] - Italic text
  * @property {boolean=} [style.underline = false] - Underlined text
- * @property {('responsive'|'fluid'|'fixed')} [sizeBehavior='responsive'] - Setting to determine how the fontsize of the label is calculated.
+ * @property {('responsive'|'relative'|'fixed')} [sizeBehavior='responsive'] - Setting to determine how the fontsize of the label is calculated.
```
haxxmaxx, line 162: Has there been a release where we use `fluid`? In that case this is a breaking change. How come we want to rename this? It still says fluid in the property panel right?

Donya-qlik: No we actually reverted it because of the issue I solved in this PR. As in kpi `fluid` has been set as `relative` Niek suggested to use the same here.

haxxmaxx: ok good, then should we change the translation as well?

Donya-qlik: `fluid` is what has been decided for it's translation, it's the same for kpi. It seems that users are supposed to see the option as 'fluid,' but in the code, it is called 'relative.' I'm agree it would be better to choose only one and keep using that everywhere but maybe in another PR after getting @niekvanstaveren's idea as well.

a-m-dev (approved):
Code is LGTM!

PR #375, opened 2023-04-11
chore: add action-button rendering tests

Donya-qlik:
New tests cover the buttons with 

- Font sizeBehavior of `relative`, `fixed` and `responsive`
- Background image
- Font family


test/rendering/__fixtures__/scenario_3.fix.js
```diff
@@ -0,0 +1,104 @@
...
+            colorExpression: '',
+            mode: 'media',
+            size: 'fitWidth',
+            position: 'top-center',
+            url: {
+              qStaticContentUrl: {
+                qUrl: '/content/default/qlik-multi-kpi-update/multikpi.png',
+              },
+            },
+          },
```
Donya-qlik, line 64: @niekvanstaveren I have some background image settings here but I don't see any sign of the it there on the button! ... either on the scenario_1 testcase which test useImage.

niekvanstaveren: Pretty sure this image is not available in this rendering test. Looks like this was copied over from the multi kpi. We should not have these options here if they are not used for the tests

niekvanstaveren (approved):
We should probably add a lot more tests here, but I can put that on the backlog.

PR #379, opened 2023-04-21
chore: use qlik prettier and lint

niekvanstaveren:
Mostly prettier changes. Will add comments to the manual changes I had to do


package.json
```diff
@@ -59,8 +53,35 @@
...
+      "no-restricted-globals": 0,
+      "no-unused-expressions": 0,
+      "object-curly-newline": 0,
+      "operator-linebreak": 0,
+      "import/no-extraneous-dependencies": [
+        2,
+        {
+          "devDependencies": true
+        }
+      ]
```
niekvanstaveren, line 83: A lot of rules here. We should check if we can remove some of these rules

src/components/__tests__/action-button.spec.js
```diff
@@ -16,152 +16,155 @@ describe('action button', () => {
...
 
-    it('should run without navigation', async () => {
+    it("should run without navigation", async () => {
       defaults.layout.navigation = {};
       renderButton(defaults);
       await defaults.element.firstElementChild.onclick();
-      expect(button.removeAttribute).toHaveBeenCalledWith('disabled');
-      expect(defaults.senseNavigation.goToSheet).toNotHaveBeenCalled;
+      expect(button.removeAttribute).toHaveBeenCalledWith("disabled");
+      expect(defaults.senseNavigation.goToSheet).toHaveBeenCalledTimes(0);
```
niekvanstaveren, line 61: Actually think this was not asserting anything at all. This should be the correct pattern

src/utils/__tests__/actions.spec.js
```diff
@@ -470,35 +470,28 @@ describe('actions', () => {
...
-    it('should call refreshDynamicViews', async () => {
+    it("should call refreshDynamicViews", async () => {
       senseNavigation.refreshDynamicViews = jest.fn();
-      const actionObject = actions.find((action) => action.value === 'refreshDynamicViews');
+      const actionObject = actions.find((action) => action.value === "refreshDynamicViews");
       await actionObject.getActionCall({ senseNavigation })();
-      expect(senseNavigation.refreshDynamicViews).toHaveBeenCalled;
-    });
-
-    it('should not call refreshDynamicViews if not implemented in sense-client', async () => {
```
niekvanstaveren, line 483: This test did not make any sense and shows that .toNotHaveBeenCalled is not a correct assertion. senseNavigation.refreshDynamicViews is not a mock. I don't think we can check this test case because the code checks if refreshDynamicViews is a function before calling it. That won't work when you have to set it to a jest mock (it will automatically be a function)

src/utils/url-utils.js
```diff
@@ -1,34 +1,33 @@
...
 
 function getSenseServerUrl(app) {
-  let config;
   let wsUrl;
   let protocol;
   let isSecure;
 
   if (app?.session?.config) {
-    config = app.session.config;
+    const { config } = app.session;
```
niekvanstaveren, line 14: Preferred to destruct

test/integration/setup.spec.js
```diff
@@ -1,4 +1,5 @@
-const serve = require('@nebula.js/cli-serve'); // eslint-disable-line
+/* eslint-disable no-undef */
```
niekvanstaveren, line 1: In a next PR I will try to tackle the integration test setup. Will try to get rid of this as well

test/integration/rendering.spec.js
```diff
@@ -1,19 +1,21 @@
+/* eslint-disable no-undef */
```
niekvanstaveren, line 1: Next PR

package.json
cbt1: Agree. The value of a shared config kind of gets diminished if that many rules are overwritten

src/ext.js
```diff
@@ -530,112 +530,112 @@ export default function ext({ translator, shouldHide, senseNavigation }) {
...
                         },
                         {
-                          translation: 'properties.dock.right',
-                          value: 'right',
+                          translation: "properties.dock.right",
+                          value: "right",
                         },
                       ],
-                      show: (data) => propertyResolver.getValue(data, 'style.icon.useIcon'),
+                      show: (data) => propertyResolver.getValue(data, "style.icon.useIcon"),
```
cbt1, line 638: What is up with the `support` declaration at line :651 ?

Also this file seems like a good candiate to follow the max length rule

```js
support: {
      export: !1,
      exportData: !1,
      snapshot: !1,
      viewData: !1,
      quickMobile: !0,
},
```

cbt1 (approved):
LGTM. I have mostly just skimmed through the changes that appear to be because of a rule change in lint/prettier

src/ext.js
niekvanstaveren: Yeah plenty of other improvements that can be made in this repo. Not sure why this is noted like this, but maybe @Caele had a good reason for this?

PR #382, opened 2023-05-02
fix: automation being triggered but not polled

rileymd88:
This fixes an issue where a user may be able to trigger an automation from the button but does not have permission to poll for its status. For this to happen the following needs to be true:

Run mode: triggered needs to be checked in the button properties
The user that clicks the button is not the owner of the automation
Start block Run Mode in the automation needs to have Run asynchronously set to true


src/utils/automation-helper.js
```diff
@@ -32,6 +32,9 @@ export const getAutomation = async (automationId) => {
 
 export const getAutomationRun = async (automationId, runId) => {
   const response = await fetch(`../api/v1/automations/${automationId}/runs/${runId}`);
+  if (response.status !== 200) {
```
cbt1, line 35: What kind of response code (status) is returned when the user does not have permissions? If 403, you should consider to only check for that

src/utils/automation-helper.js
```diff
@@ -32,6 +32,9 @@ export const getAutomation = async (automationId) => {
...
+    throw Error;
```
cbt1, line 36: You are throwing the `Error` function here. Sort of like `throw () => {}`.

Consider something like `throw new Error()` instead.

rileymd88: Fixed this in the latest commit

cbt1 (approved):

cbt1:
Fix the failing unit tests and you are good to go

rileymd88:
> Fix the failing unit tests and you are good to go

Done in the latest commit :)

niekvanstaveren:
If still valid this needs to be recreated from a direct branch, not from a Fork

PR #384, opened 2023-05-04
fix: fluid and fix sizeBehavior with italic style might cut the label

Donya-qlik:
**Issue:** 
- selecting fixed and fluid font sizeBehavior + Italic font style cuts the button label at the end for some specific font families.
- `stretch to fit` size is missed so the background-image size can't adjust size to the button size.

**Solution:** 
- comparing the label and button widths and set them to visible as long as the label's width is smaller than the button's width.
- Adding StretchFit to the background size list and set it to 100%

The PR also covers a new design decision to set the font sizeBehavior default value to `fixed`.


api-specifications/properties.json
```diff
@@ -615,7 +615,7 @@
         "sizeBehavior": {
           "description": "Setting to determine how the fontsize of the label is calculated.",
           "optional": true,
-          "defaultValue": "'responsive'",
+          "defaultValue": "fixed",
```
eliseeborn, line 618:
```suggestion
          "defaultValue": "'fixed'",
```
the double quotation marks are strange, but I think they are needed

src/object-properties.js
```diff
@@ -175,7 +175,7 @@ const properties = {
  * @property {object=} url.qStaticContentUrlDef
  * @property {string=} url.qStaticContentUrlDef.qUrl - URL represented as a string
  * @property {('topLeft'|'centerLeft'|'bottomLeft'|'topCenter'|'centerCenter'|'bottomCenter'|'topRight'|'centerRight'|'bottomRight'|'top-left'|'center-left'|'bottom-left'|'top-center'|'center-center'|'top-right'|'center-right'|'bottom-right')} [position='center-center'] - Image position
- * @property {('auto'|'alwaysFit'|'fitWidth'|'fitWidth'|'fitHeight'|'fill'|'alwaysFill')=} [size='auto'] - Size of the image, relative to the button
+ * @property {('auto'|'alwaysFit'|'fitWidth'|'fitWidth'|'fitHeight'|'fill'|'alwaysFill'|'stretchFit')=} [size='auto'] - Size of the image, relative to the button
```
eliseeborn, line 178: there are two fitWidth here
```suggestion
 * @property {('auto'|'alwaysFit'|'fitWidth'|'fitHeight'|'fill'|'alwaysFill'|'stretchFit')=} [size='auto'] - Size of the image, relative to the button
```

eliseeborn, line 178: there are two fitWidth here
```suggestion
 * @property {('auto'|'alwaysFit'|'fitWidth'|'fitHeight'|'fill'|'alwaysFill'|'stretchFit')=} [size='auto'] - Size of the image, relative to the button
```

eliseeborn (approved):
just a few quick fixes

PR #394, opened 2023-05-22
chore: set responsive as sizeBehavior default value

Donya-qlik:
It has been decided to turn back the sizeBehavior default value of sn-action-button to `responsive`.


Donya-qlik:
The api-governance request to change the properties is in progress, but it will not have any effect on the applied changes on this PR.

sharleneaqlik (approved):
LGTM! though I haven't tested manually.

niekvanstaveren (approved):

PR #395, opened 2023-05-23
fix: handle error when migrating automation action

rileymd88:
This fixes an issue where the automation action does not have an `automationId` property set up yet which can cause the automation action to fail


src/utils/automation-props.js
```diff
@@ -44,7 +44,8 @@ const getAutomationProps = (multiUserAutomation, getAutomations) => ({
     translation: "Object.ActionButton.Automation.Link",
     component: "link",
     url: (data) => `/automations/editor/${data.automationId}`,
-    show: (data) => data.automationId.length > 1 && multiUserAutomation,
+    show: (data) =>
+      Object.prototype.hasOwnProperty.call(data, "automationId") && data.automationId.length > 1 && multiUserAutomation,
```
niekvanstaveren, line 48: Can't you just check existence of automationId with ?

niekvanstaveren (commented):
Is this still an issue?

rileymd88:
> Is this still an issue?

Yes

src/utils/automation-props.js
rileymd88: I will update this

niekvanstaveren (approved):

PR #396, opened 2023-05-24
fix: make sure bookmark layout gets applied to app

AmeliaAndersson:
Before when activating a bookmark through action button, a layout wouldnt be applied if there was one before. Now we check if the bookmark we want to use has a layout and if it does we remove all previous layout so the new one can be applied.

Tested on localhost.

### DIFF
--- a/src/utils/actions.js
+++ b/src/utils/actions.js
@@ -33,6 +33,10 @@ const actions = [
       ({ app, bookmark }) =>
       async () => {
         const bookMarks = await app.getBookmarkList();
+        const findBmWithID = bookMarks.find((bm) => bm.qInfo.qId === bookmark);
+        if (findBmWithID.qData.qBookmark.qPatches.length > 0 || findBmWithID.qMeta.isExtended) {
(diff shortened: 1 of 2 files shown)


src/utils/actions.js
```diff
@@ -33,6 +33,10 @@ const actions = [
       ({ app, bookmark }) =>
       async () => {
         const bookMarks = await app.getBookmarkList();
+        const findBmWithID = bookMarks.find((bm) => bm.qInfo.qId === bookmark);
+        if (findBmWithID.qData.qBookmark.qPatches.length > 0 || findBmWithID.qMeta.isExtended) {
+          app.clearAllSoftPatches();
```
niekvanstaveren, line 38:
```suggestion
          app.clearAllSoftPatches?.();
```

I think like this

niekvanstaveren (approved):

niekvanstaveren:
Code looks okay. Need to update some of the tests though. Api-governance should also work again after another run

PR #398, opened 2023-05-24
fix: insert whitespace changes the font size

Donya-qlik:


Donya-qlik:
Moved it to: https://github.com/qlik-oss/sn-action-button/pull/400

PR #402, opened 2023-06-01
fix: providing feature flag to font resolver in qlik-chart-modules

StefanEnberg:
providing feature flag to font resolver in qlik-chart-modules


src/index.js
```diff
@@ -75,6 +75,6 @@ export default function supernova(env) {
         [element]
       );
     },
-    ext: ext({ translator, shouldHide, senseNavigation }),
+    ext: ext({ translator, shouldHide, senseNavigation, ext, theme: sense?.theme }),
```
cbt1, line 78: You are passing the imported `ext` function as a property here.

Also `env.sense.theme` is deprecated. Use `env.anything.sense.theme` instead.

cbt1 (approved):
LGTM

PR #404, opened 2023-06-16
fix: remove unsupported softLock actions

johanlahti:
Remove actions which do no work with `softLock` (the checkbox label suggests that they should overwrite locked fields, but this does not work, which is also in accordance with Engine design).

I have gone through all actions and found that it works only for the following actions. For others I have removed the checkbox:
- clearAll
- selectAll
- selectAlternative
- selectExcluded
- selectValues

The same has been done in sense-client for sheet actions and this PR will make them aligned.

### Before:

image

### After (soft lock removed):

image

### DIFF
--- a/src/utils/actions.js
+++ b/src/utils/actions.js
@@ -80,14 +80,15 @@ const actions = [
     translation: "Object.ActionButton.ClearAllButThis",
     group: "selection",
     getActionCall:
-      ({ app, qStateName, field, softLock }) =>
+      ({ app, qStateName, field }) =>
       async () => {
         if (field) {
           const fieldObj = await app.getField(field, qStateName);
+          const softLock = false;
           await fieldObj.clearAllButThis(softLock);
         }
       },
-    requiredInput: ["field", "softLock"],
+    requiredInput: ["field"],
   },
   {
     value: "clearField",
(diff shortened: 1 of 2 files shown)


haxxmaxx:
We probably need to change here https://help.qlik.com/en-US/cloud-services/Subsystems/Hub/Content/Sense_Hub/Visualizations/Button/create-buttons.htm

johanlahti:
@haxxmaxx I have contacted @tklkan to update the docs. To be completely aligned, I guess we should do this when this has been merged, released and bumped in sense-client – but since the functionality is broken, it's "almost ok" if it's removed from the docs before that's done.

src/utils/actions.js
```diff
@@ -138,14 +139,15 @@ const actions = [
     translation: "Object.ActionButton.SelectMatchingValues",
     group: "selection",
     getActionCall:
-      ({ app, qStateName, field, value, softLock }) =>
+      ({ app, qStateName, field, value }) =>
       async () => {
         if (field && value) {
           const fieldObj = await app.getField(field, qStateName);
+          const softLock = false;
           await fieldObj.select(value, false, softLock);
```
haxxmaxx, line 147: If the softLock arg doesn't do anything, why do we have to pass it? could we just pass two arguments?

haxxmaxx (approved):

PR #406, opened 2023-06-30
ci: move to sn-gh-workflows

niekvanstaveren:


package.json
```diff
@@ -37,9 +37,9 @@
     "sense": "nebula sense",
     "spec": "scriptappy-from-jsdoc -c ./spec-configs/props.conf.js",
     "start": "nebula serve",
+    "test:e2e": "playwright test",
```
haxxmaxx, line 40: the rendering tests became e2e tests?

niekvanstaveren: The default for nebula charts is e2e tests https://github.com/qlik-oss/nebula.js/blob/master/commands/create/templates/sn/none/_package.json#L22

We have not really figured out if we need to split them up or not in the script. I think both rendering and integration tests can both live under the test:e2e command

haxxmaxx: ok makes sense

haxxmaxx (approved):

niekvanstaveren:
We can only fully move when sn-gh-workflows support API governance through the release process. Will come back to this after my holiday

PR #412, opened 2023-08-03
fix: QB-20783-missing space between icon and label on the button object

sharleneaqlik:
On the button object when the "layout behaviour" is selected as Fixed or Fluid along with icon then there is a space missing as shown in the screenshot below. This PR fixes that issue. Although the bug is for fluid only but this PR fixes for both selections (fixed and fluid). 

image

Fixed behaviour: 

image


src/utils/style-utils.js
```diff
@@ -90,6 +90,7 @@ export const setTextFontSize = (text, font, textFontSize, hasIcon) => {
 
 const setFontSizeBehaviorStyle = (button, text, textSpan) => {
   text.offsetWidth <= button.clientWidth ? (textSpan.style.overflow = "visible") : (textSpan.style.overflow = "hidden");
+  text.children[0].style.marginRight = `${trimDecimal(text.offsetWidth * 0.04)}px`;
```
niekvanstaveren, line 93: What happens when there is no icon selected?

sharleneaqlik: It is displayed this way when no icon set. 
image

niekvanstaveren: But it sets a marginRight on the text instead then? I guess that is why all the rendering tests are complaining

sharleneaqlik: It should not, because it only sets the margin for the first child which is the icon.

niekvanstaveren: Yes, but when you do not have an icon selected the first child would be the text instead right?

sharleneaqlik: Indeed, a hasIcon check should be there. REverting the rendering tests first and then adding the check.

eliseeborn: also, isn't there a setting to have the icon to the right of the text instead of to the left?

sharleneaqlik: Will check those options too. Thanks for pointing out.

sharleneaqlik: @eliseeborn , checked all the combo. Seems work. Now need to see if rendering tests need fix, since there has been a little UI change with the space. 
image

eliseeborn (approved):

PR #421, opened 2023-08-29
fix: button background image is missing in the downloaded sheet

Donya-qlik:
**Issue:** 

- Downloading sheet includes button with background image, results a button without background image in preview window

**Solution:**

- Defining a promise to wait for the image to be loaded in the background before setting the image url as the background-image. It prevents the button from reporting to the reporting service while the image is loading.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,7 @@
     "@playwright/test": "1.37.1",
     "@qlik/eslint-config": "0.4.17",
     "@qlik/prettier-config": "0.3.0",
+    "@testing-library/react": "^14.0.0",
     "babel-plugin-istanbul": "6.1.1",
     "eslint": "8.49.0",
     "fs-extra": "11.1.1",
@@ -102,6 +103,8 @@
     "mocha-junit-reporter": "2.2.1",
     "prettier": "2.8.8",
     "qlik-chart-modules": "0.59.0",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
     "scriptappy-from-jsdoc": "0.7.0",
     "shx": "0.3.4",
     "yargs": "17.7.2"
@@ -115,10 +118,5 @@
   "publishConfig": {
     "access": "public"
   },
-  "systemjs": "dist/sn-action-button.systemjs.js",
-  "dependencies": {
-    "@testing-library/react": "^14.0.0",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0"
-  }
+  "systemjs": "dist/sn-action-button.systemjs.js"
 }
--- a/package.json
+++ b/package.json
@@ -91,7 +91,6 @@
     "@playwright/test": "1.37.1",
     "@qlik/eslint-config": "0.4.17",
     "@qlik/prettier-config": "0.3.0",
-    "@testing-library/react": "^14.0.0",
     "babel-plugin-istanbul": "6.1.1",
     "eslint": "8.49.0",
     "fs-extra": "11.1.1",
@@ -103,8 +102,6 @@
     "mocha-junit-reporter": "2.2.1",
     "prettier": "2.8.8",
     "qlik-chart-modules": "0.59.0",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0",
     "scriptappy-from-jsdoc": "0.7.0",
     "shx": "0.3.4",
     "yargs": "17.7.2"
--- a/package.json
+++ b/package.json
@@ -115,5 +115,10 @@
   "publishConfig": {
     "access": "public"
   },
-  "systemjs": "dist/sn-action-button.systemjs.js"
+  "systemjs": "dist/sn-action-button.systemjs.js",
+  "dependencies": {
+    "@testing-library/react": "^14.0.0",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
+  }
 }
--- a/src/__tests__/index.spec.js
+++ b/src/__tests__/index.spec.js
@@ -35,7 +35,7 @@ describe("index", () => {
     },
   };
 
-  it("should render supernova", async () => {
+  it("should render supernova twice", async () => {
     const translator = { get: () => "fakeTranslation" };
     const result = supernova({
       anything: {
@@ -55,27 +55,23 @@ describe("index", () => {
           background: {
             url: {
               aStaticContentUrl: {
-                qUrl: "",
+                qUrl: "/media/Logo/qlik.png",
               },
             },
           },
         },
       },
       constraints: "constraints",
-      imageUrl: null,
     });
 
-    await c.update();
-
-    expect(thisElement.appendChild).toHaveBeenCalled();
-    expect(renderSpy).toHaveBeenCalledWith({
+    const value = {
       element: thisElement,
       layout: {
         style: {
           background: {
             url: {
               aStaticContentUrl: {
-                qUrl: "",
+                qUrl: "/media/Logo/qlik.png",
               },
             },
           },
@@ -87,7 +83,11 @@ describe("index", () => {
       senseNavigation: "nav",
       multiUserAutomation: true,
       translator,
-      imageUrl: null,
-    });
+    };
+
+    await c.update();
+
+    expect(thisElement.appendChild).toHaveBeenCalled();
+    expect(renderSpy).toHaveBeenNthCalledWith(2, value);
   });
 });
--- a/src/__tests__/index.spec.js
+++ b/src/__tests__/index.spec.js
@@ -50,8 +50,19 @@ describe("index", () => {
     });
     const c = create(result.component, {
       element: thisElement,
-      layout: "layout",
+      layout: {
+        style: {
+          background: {
+            url: {
+              aStaticContentUrl: {
+                qUrl: "",
+              },
+            },
+          },
+        },
+      },
       constraints: "constraints",
+      imageUrl: null,
     });
 
     await c.update();
@@ -59,13 +70,24 @@ describe("index", () => {
     expect(thisElement.appendChild).toHaveBeenCalled();
     expect(renderSpy).toHaveBeenCalledWith({
       element: thisElement,
-      layout: "layout",
+      layout: {
+        style: {
+          background: {
+            url: {
+              aStaticContentUrl: {
+                qUrl: "",
+              },
+            },
+          },
+        },
+      },
       app: undefined,
       constraints: "constraints",
       theme: undefined,
       senseNavigation: "nav",
       multiUserAutomation: true,
       translator,
+      imageUrl: null,
     });
   });
 });
--- a/src/hooks/use-load-image.js
+++ b/src/hooks/use-load-image.js
@@ -20,7 +20,7 @@ const useLoadImage = (layout, app) => {
       return imgSrc.src;
     }
     return null;
-  }, [layout, app]);
+  }, [url, app]);
   return imageUrl;
 };
 
--- a/src/hooks/use-load-image.js
+++ b/src/hooks/use-load-image.js
@@ -20,7 +20,7 @@ const useLoadImage = (layout, app) => {
       return imgSrc.src;
     }
     return null;
-  }, [layout]);
+  }, [layout, app]);
   return imageUrl;
 };
 
--- a/src/hooks/use-load-image.js
+++ b/src/hooks/use-load-image.js
@@ -2,34 +2,26 @@ import { usePromise } from "@nebula.js/stardust";
 import { getImageUrl } from "../utils/url-utils";
 
 // Create a promise that ensures the button background image is loaded properly
+// and the image url is resolved.
 const imagePromise = (url, app) =>
   new Promise((resolve, reject) => {
     const imgUrl = getImageUrl(url, app);
     const img = new Image();
     img.src = imgUrl;
     img.onload = () => resolve(img);
-    img.onerror = () => reject(img);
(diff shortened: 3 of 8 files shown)


src/index.js
```diff
@@ -57,6 +59,7 @@ export default function supernova(env) {
         senseNavigation,
         multiUserAutomation,
         translator,
+        isBackgroundLoaded,
```
haxxmaxx, line 62: We should probably wrap `renderButton` in a `useEffect` and add `isBackgroundLoaded` to the dependency array. Then make sure `isBackgroundLoaded` is true, to run `renderButton`.

src/hooks/use-load-image.js
```diff
@@ -0,0 +1,33 @@
...
+    const img = new Image();
+    img.src = url;
+    img.onload = () => resolve(img);
+    img.onerror = () => reject(img);
+  });
+
+const useLoadImage = (layout) => {
+  const { url } = layout.style.background;
+  const [isBackgroundLoaded = false, error] = usePromise(async () => {
+    if (url && url.qStaticContentUrl && url.qStaticContentUrl.qUrl) {
```
haxxmaxx, line 15:
```suggestion
    if (url?.qStaticContentUrl?.qUrl) {
```

src/hooks/use-load-image.js
```diff
@@ -0,0 +1,33 @@
+import { usePromise } from "@nebula.js/stardust";
+
+// Create a promise that ensures the button background image is loaded properly
+const imagePromise = (url) =>
+  new Promise((resolve, reject) => {
+    const img = new Image();
+    img.src = url;
```
haxxmaxx, line 7: In style formatter uses `getImageUrl` to edit the url, do you need to do similar here?

src/hooks/use-load-image.js
```diff
@@ -0,0 +1,33 @@
...
+      try {
+        await imagePromise(url.qStaticContentUrl.qUrl);
+        return true;
+      } catch {
+        return false;
+      }
+    }
+    return false;
+  }, [layout]);
+  if (error) {
```
haxxmaxx, line 25: you are catching the possible error from `imagePromise`, I don't think you get one from `usePromise`.

haxxmaxx (commented):
This might be a viable approach, would like some more eyes on it

src/index.js
Donya-qlik: Why do you think  it is needed ?
`isBackgroundLoaded ` is called when the url changes in the layout via `useLoadImage`. It keeps the button rendering until the image is loaded properly. I don't think I need to change anything else.

src/hooks/use-load-image.js
Donya-qlik: Do you mean returning `false` ? fixed it.

src/index.js
haxxmaxx: Because it is kind of a backwards way of doing it IMO. You shouldn't really render anything until you actually have loaded the image. Then you don't have to pass a boolean all the way to styling logic

src/hooks/use-load-image.js
haxxmaxx: No, I mean that you are basically handling the error twice basically. One try catch block inside the usePromise, and then checking the error outside usePromise as well. Not sure, but I think that usePromise will never return an error since you catch it in usePromise

src/index.js
Donya-qlik: `isBackgrounfLoaded` is a check for the case that the button gets background image, it can be true or false. When the button has no background image the url is undefined and we still need the button to be rendered and get the rest of the style.

Donya-qlik:
@haxxmaxx I retested my solution and seems the bug is still there! I was careful to keep the testing environment clean for each test and I couldn't reproduce it anymore for at least for a day but yesterday it showed up again! :(  Not sure what might be the case, I think the solution waits enough for the image to be loaded.

src/index.js
haxxmaxx: Then we can just always make isBackgroundLoaded to be true when there is no background

haxxmaxx:
> @haxxmaxx I retested my solution and seems the bug is still there! I was careful to keep the testing environment clean for each test and I couldn't reproduce it anymore for at least for a day but yesterday it showed up again! :( Not sure what might be the case, I think the solution waits enough for the image to be loaded.

Yeah there probably needs to be something on the downloading side waiting for the button to finish rendering. Not sure how other charts do that actually.

niekvanstaveren (commented):
Not completely sure about this approach. We don't need to pre-load the image, that can just be loaded like it was before. The only thing we need is to return in the usePromise when the image is loaded.

package.json
```diff
@@ -115,5 +115,10 @@
   "publishConfig": {
     "access": "public"
   },
-  "systemjs": "dist/sn-action-button.systemjs.js"
+  "systemjs": "dist/sn-action-button.systemjs.js",
+  "dependencies": {
+    "@testing-library/react": "^14.0.0",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
+  }
```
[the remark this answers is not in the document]
Donya-qlik: I added them because of the unit tests, to use (WaitFor)

src/hooks/use-load-image.js
Donya-qlik: Ah, true! will remove it.

src/index.js
Donya-qlik: I don't know which solution is correct. The bug is still there,  I could reproduce it however it happens less often. I'll try your idea for not passing the variable down into getStyles.

package.json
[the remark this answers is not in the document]
cbt1: If they are really needed they should be devDependencies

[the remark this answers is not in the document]
Donya-qlik: Well, you know they are not something we add manually, when I installed testing-library/react and the others, they were just automatically added. Is there a way to control where they be added ?

package.json
```diff
@@ -91,6 +91,7 @@
     "@playwright/test": "1.37.1",
     "@qlik/eslint-config": "0.4.17",
     "@qlik/prettier-config": "0.3.0",
+    "@testing-library/react": "^14.0.0",
```
cbt1, line 94: The versions should be pinned. Also don't forget to run `yarn` otherwise the lock file won't have pinned versions.

src/hooks/use-load-image.js
```diff
@@ -0,0 +1,27 @@
...
+
+const useLoadImage = (layout, app) => {
+  const { url } = layout.style.background;
+  const [imageUrl] = usePromise(async () => {
+    if (url?.qStaticContentUrl?.qUrl) {
+      const imgSrc = await imagePromise(url.qStaticContentUrl.qUrl, app);
+      return imgSrc.src;
+    }
+    return null;
+  }, [layout]);
```
cbt1, line 23: `layout` is not being used inside the hook. So no point any "reloading" the image every time the layout changes.

Also not sure why you don't get a lint error about the missing `app` dependency.

```suggestion
  }, [url, app]);
```

src/utils/__tests__/style-formatter.spec.js
```diff
@@ -31,133 +34,145 @@ describe("style-formatter", () => {
...
+    it("should set the image url when the image url exists", async () => {
       style.background.useImage = false;
       style.background.mode = "media";
       style.background.url.qStaticContentUrl = { qUrl: someUrl };
-      const formattedStyle = styleFormatter.getStyles({ style, disabled, theme, app });
-      expect(formattedStyle.includes(`background-image: url('https://localhost${someUrl}')`)).toBe(true);
+      imageUrl = await new Promise((resolve) => {
+        setTimeout(() => resolve(`https://localhost${someUrl}`), 1000);
+      });
+      await waitFor(() => expect(imageUrl).toBe(`https://localhost${someUrl}`));
```
cbt1, line 114: This is just verifying the promise that was created on the previous line.

What is that you really want to test here?

src/utils/__tests__/style-formatter.spec.js
```diff
@@ -31,133 +34,145 @@ describe("style-formatter", () => {
...
-      const formattedStyle = styleFormatter.getStyles({ style, disabled, theme, app });
-      expect(formattedStyle.includes("background-image:")).toBe(false);
-      expect(formattedStyle.includes("background-size:")).toBe(false);
-      expect(formattedStyle.includes("background-position:")).toBe(false);
-      expect(formattedStyle.includes("background-repeat:")).toBe(false);
+      style.background.url.qStaticContentUrl = { qUrl: someUrl };
+      imageUrl = await new Promise((reject) => {
+        setTimeout(() => reject(undefined), 500);
+      });
+      await waitFor(() => expect(imageUrl).toBe(undefined));
```
cbt1, line 129: This is just verifying the promise that was created on the previous line.

What is that you really want to test here?

src/utils/__tests__/style-formatter.spec.js
```diff
@@ -31,133 +34,145 @@ describe("style-formatter", () => {
...
 
-    it("should return no settings when url is missing", () => {
+    it("should return no settings when url is missing", async () => {
       style.background.useImage = true;
-      style.background.url.qStaticContentUrl = {};
-      const formattedStyle = styleFormatter.getStyles({ style, disabled, theme, app });
+      imageUrl = await new Promise((reject) => {
+        setTimeout(() => reject(undefined), 500);
+      });
+      await waitFor(() => expect(imageUrl).toBe(undefined));
```
cbt1, line 174: This is just verifying the promise that was created on the previous line.

What is that you really want to test here?

Donya-qlik: Yes waitFor line is just for verifying the promise, resolve/reject is returned. the line after that uses imageUrl and the rest of scenario. I thought it good to have it.

src/hooks/use-load-image.js
Donya-qlik: I used layout at line 16 and is used in `imagePromise`. I fixed the missing app.

cbt1: Yes but that is outside of the `usePromise` hook?

src/utils/__tests__/style-formatter.spec.js
cbt1: But in case `imageUrl` is always going to be `undefined`. So wrapping it in a promise does not change anything. All it does is make the test case 500 ms longer.

Can get the exact same test by doing `imageUrl = undefined`, just that is going to be a lot faster.

src/utils/__tests__/style-formatter.spec.js
```diff
@@ -31,133 +34,145 @@ describe("style-formatter", () => {
...
       expect(formattedStyle.includes("background-repeat: no-repeat")).toBe(true);
     });
 
-    it("should return no settings when url is missing", () => {
+    it("should return no settings when url is missing", async () => {
       style.background.useImage = true;
-      style.background.url.qStaticContentUrl = {};
-      const formattedStyle = styleFormatter.getStyles({ style, disabled, theme, app });
+      imageUrl = await new Promise((reject) => {
+        setTimeout(() => reject(undefined), 500);
```
cbt1, line 172: `reject` is actually the `resolve` callback

src/utils/__tests__/style-formatter.spec.js
Donya-qlik: True, it can only be undefined without any timeout. I've already verified that imgeUrl waits for loading anyway no need for reject case. I'll fix it.

src/hooks/use-load-image.js
Donya-qlik: I'm not sure how the hook counts it but seems putting url affects result the same. I'll fix it.

src/utils/__tests__/style-formatter.spec.js
cbt1: Just assign the url directly instead of waiting for it to be done `imageUrl = `https://localhost${someUrl}`

src/utils/__tests__/style-formatter.spec.js
```diff
@@ -31,133 +33,138 @@ describe("style-formatter", () => {
...
+      imageUrl = `https://localhost${someUrl}`;
     });
 
     afterEach(() => {
       jest.resetAllMocks();
     });
 
     it("should return default styling", () => {
-      const formattedStyle = styleFormatter.getStyles({ style, disabled, theme, app });
+      const formattedStyle = styleFormatter.getStyles({ style, disabled, theme, app, imageUrl });
```
niekvanstaveren, line 44: app was removed as input parameter so you can remove that here as well

niekvanstaveren (changes requested):
The url to the image is now only resolved through a promise. However there is no need for that. Getting the URL is all synchronous, it is only the loading of the image itself that is asynchronous. 
Makes parts of the code unnecessary complex.

src/__tests__/index.spec.js
```diff
@@ -50,22 +50,44 @@ describe("index", () => {
...
       senseNavigation: "nav",
       multiUserAutomation: true,
       translator,
-    });
+    };
+
+    await c.update();
+
+    expect(thisElement.appendChild).toHaveBeenCalled();
+    expect(renderSpy).toHaveBeenNthCalledWith(2, value);
```
niekvanstaveren, line 91: This is a bit weird that it would render twice now I think. When looking in the nebula dev environment I still see the render count as 1, so I guess it is only in the unit test situation

niekvanstaveren (approved):

PR #428, opened 2023-09-18
chore: update scriptappy

niekvanstaveren:


spec-configs/props.conf.js
```diff
@@ -3,24 +3,28 @@ const path = require("path");
...
-  },
-  parse: {
-    types: {
-      "qae.GenericObjectProperties": {},
+    output: {
+      file: path.resolve(__dirname, "../api-specifications/properties.json"),
+    },
+    parse: {
+      types: {
+        GenericObjectProperties: {
```
Caele, line 24:
```suggestion
        EngineAPI.GenericObjectProperties: {
```

src/object-properties.js
```diff
@@ -1,7 +1,7 @@
 import DEFAULTS from "./style-defaults";
 
 /**
- * @extends {qae.GenericObjectProperties}
+ * @extends {GenericObjectProperties}
```
Caele, line 4:
```suggestion
 * @extends {EngineAPI.GenericObjectProperties}
```

spec-configs/props.conf.js
Caele: + update in the props file and run it again

niekvanstaveren: But the we should also reference it like that here https://github.com/qlik-oss/sn-action-button/pull/428

niekvanstaveren: Also how wordcloud actually references it https://github.com/qlik-oss/sn-word-cloud/blob/main/api-spec/props.conf.js#L26

package.json
```diff
@@ -89,6 +89,8 @@
     "@playwright/test": "1.38.0",
     "@qlik/eslint-config": "0.4.18",
     "@qlik/prettier-config": "0.3.0",
+    "@scriptappy/cli": "^0.10.0",
```
cbt1, line 92: Should pin the dependencies

cbt1 (approved):
Cannot really comment on how valid the spec changes are but it LGTM

niekvanstaveren:
> Cannot really comment on how valid the spec changes are but it LGTM

No it is quite a diff. Tried to compare it and it looks mostly like re-ordering

PR #429, opened 2023-09-19
chore: cleanup circle config

niekvanstaveren:


cbt1 (approved):
I think you will need to "remove" the project from circleci as well

niekvanstaveren:
> I think you will need to "remove" the project from circleci as well

I blocked builds right now. Maybe good to keep around in case we need to build old release tracks

PR #434, opened 2023-10-02
chore: typescript setup

Donya-qlik:


jest.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  testEnvironment: "jsdom",
+  testEnvironment: "jest-environment-jsdom",
```
Donya-qlik, line 2: It's required for jest> 28

package.json
```diff
@@ -73,6 +73,12 @@
       ]
     }
   },
+  "dependencies": {
+    "@types/jest": "^29.5.5",
+    "@types/node": "^20.7.1",
+    "jest-environment-jsdom": "^29.7.0",
+    "typescript": "^5.2.2"
```
niekvanstaveren, line 80: These should be dev dependencies and locked

tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "compilerOptions": {
+    "jsx": "react",
```
niekvanstaveren, line 3: There is no react

tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
```
niekvanstaveren, line 1: Can we extend qlik tsconfig? It is added as dependency...

package.json
```diff
@@ -86,17 +92,17 @@
     "@nebula.js/cli-serve": "4.4.0",
     "@nebula.js/stardust": "4.4.0",
     "@nebula.js/test-utils": "0.6.0",
-    "@playwright/test": "1.38.1",
-    "@qlik/eslint-config": "0.4.21",
-    "@qlik/prettier-config": "0.4.0",
+    "@playwright/test": "1.38.0",
```
niekvanstaveren, line 95: This is a downgrade...

package.json
```diff
@@ -86,17 +92,17 @@
...
+    "@qlik/eslint-config": "0.4.18",
+    "@qlik/prettier-config": "0.3.0",
```
niekvanstaveren, line 97: This is a downgrade. A specific reason here?

niekvanstaveren (changes requested):

niekvanstaveren:
To fully test this you will need to convert a file to typescript as well. Also a script to check the types would be good to have in that case

package.json
Donya-qlik: Moved them to devDep

Donya-qlik:
> To fully test this you will need to convert a file to typescript as well. Also a script to check the types would be good to have in that case

Should it be a specific file for this? 
what kind of script do you mean ? unit test ?

niekvanstaveren:
> > To fully test this you will need to convert a file to typescript as well. Also a script to check the types would be good to have in that case
> 
> Should it be a specific file for this? what kind of script do you mean ? unit test ?

No it can probably be any file. Look for something that could really benefit from being typed I guess.
I am talking about a script in package.json. If you look at other repos they will have an entry:

`types:check: "tsc --noEmit"` (Or something similar)

Donya-qlik:
I decided to close this PR temporarily until the refactoring [PR](https://github.com/qlik-oss/sn-action-button/pull/436) is merged

PR #435, opened 2023-10-04
feat: IM-4975-action-button add action to go to chart

sharleneaqlik:
This is a Draft PR to add an action, to the action button that takes the user to a sheet with a specified chart in focus. Not clarified yet, should it be Action or Navigation(please do find out). 

UserFlow: 

1. User clicks on the Actions Navigations for the button
2. Select the Navigation "Go to a sheet"
3. Selects which sheet to go to
4. User is then presented with the options available for all the objects on that chart
5. User selects the specific chart

Result: Clicking on the button will take user to any sheet's any objects(charts) user selects.

Initially for investigation I started looking into generating all the objects for a specific chart. Which works as 

image

However, navigating to that specific chart has not been done yet.


Donya-qlik:
The final improvement can be found here: https://github.com/qlik-oss/sn-action-button/pull/451. I'll close this PR.

PR #436, opened 2023-10-05
refactor: style-formatter

Donya-qlik:


src/utils/style-formatter.js
```diff
@@ -13,43 +19,10 @@ export default {
...
-        "border-radius",
-        `${((border.radius || DEFAULTS.BORDER_RADIUS) * lengthShortSide) / 2}px`
-      );
-    } else {
-      styles += "border: none;";
-    }
-
+    styles += getFontStyle(font, DEFAULTS.FONT_COLOR.color, theme, formatProperty);
+    styles += formatProperty("background-color", getBackgroundColor(background, theme));
+    styles += getBackgroundImage(background, app, formatProperty);
```
cbt1, line 24: This would append `undefined` to the `styles` string when `background` is falsy.

src/utils/style-formatter.js
```diff
@@ -13,43 +19,10 @@ export default {
...
-      styles += formatProperty("border", `${borderSize}px solid ${borderColor}`);
-      styles += formatProperty(
-        "border-radius",
-        `${((border.radius || DEFAULTS.BORDER_RADIUS) * lengthShortSide) / 2}px`
-      );
-    } else {
-      styles += "border: none;";
-    }
-
+    styles += getFontStyle(font, DEFAULTS.FONT_COLOR.color, theme, formatProperty);
```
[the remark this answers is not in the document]
Donya-qlik: true.. it actually comes from this line ` const { font = {}, background = {}, border = {} } = style;` that I'm not sure why we set the initial values to undefined. Their properties have always values or defaults values. I think it's better I get rid of them.

src/utils/style-formatter.js
```diff
@@ -1,55 +1,28 @@
...
+} from "./style-utils";
 
 const formatProperty = (path, setting) => `${path}: ${setting};`;
 
 export default {
-  getStyles({ style = {}, disabled, theme, element, app }) {
+  getStyles({ style, disabled, theme, element, app }) {
     let styles =
       "width: 100%;height: 100%;transition: transform .1s ease-in-out;position: absolute;bottom: 0;left: 0; top: 0;right: 0;margin: auto;";
-    const { font = {}, background = {}, border = {} } = style;
```
Donya-qlik, line 13: Having or not having these initial values has no effect on button styling values

cbt1: It might actually be needed. For example the test below will fail on your branch but works fine on`master`.

```js
    it("should return default with empty style object", () => {
      const formattedStyle = styleFormatter.getStyles({ style: {}, disabled, theme, element, app });
      expect(formattedStyle).toEqual(defaultStyle);
    });
```

Donya-qlik: ... but we have `style `in all possible scenarios, it is never empty or undefined. When the button styles reset via resetAll, we will have some empty properties like:
`style: {
  background: {useImage: false},
  font: {},
  border: {},
  ...
}` 
but the style itself stays with the main properties. This is the reason I think there is no need of them.

niekvanstaveren: What about very old buttons. Could it be that these properties were added at a later stage?

Donya-qlik: There will be no issue for those either. For old buttons style properties have also been defined in `object-properties.js` and even by resetting, style will not turn undefined. The same thing happens there, the properties might get empty but by `getStyle` at each section they get a default value in the case they are not defined.

cbt1: What about a mashup where you do something like this:
```js
const nuked = window.stardust.embed(app, {
      types: [
        {
          name: "action-button",
          load: () => Promise.resolve(window["sn-action-button"]),
        },
      ],
    });

    nuked.selections().then((s) => s.mount(document.querySelector(".toolbar")));

    // create a session object
    nuked.render({
      type: "action-button",
      element: document.querySelector(".object"),
      properties: {
        actions: [
          {
            actionType: "selectValues",
            field: "Region",
            value: "Americas;Asia",
            softLock: false,
          },
        ],
        style: {}, // or undefined
      },
    });
```

Donya-qlik: ... hadn't thought about mashups 🤔 Umm... then I'll keep them, it seems safer to have them

niekvanstaveren: > There will be no issue for those either. For old buttons style properties have also been defined in `object-properties.js` and even by resetting, style will not turn undefined. The same thing happens there, the properties might get empty but by `getStyle` at each section they get a default value in the case they are not defined.

That is not true. The object-properties are used when the button is created the first time. So if those properties did not exist at that time the button will not have them

cbt1 (approved):
LGTM

PR #439, opened 2023-10-11
feat: extend gotoSheet navigation by adding chartId option

Donya-qlik:
Adding an extra expression field to gotoSheet navigation to specify a chart in the target sheet.
The related PR of sense-client can be found on: https://github.com/qlik-trial/sense-client/pull/25579

### DIFF
--- a/src/ext.js
+++ b/src/ext.js
@@ -201,6 +201,13 @@ export default function ext({ translator, shouldHide, senseNavigation, theme }) 
                     }));
                   },
                 },
+                chart: {
+                  type: "string",
+                  expression: "optional",
+                  ref: "navigation.chart",
+                  translation: "properties.chartPicker.chart",
+                  show: (data) => checkShowNavigation(data, "sheet", "chart"),
+                },
                 story: {
                   type: "string",
                   ref: "navigation.story",
--- a/src/ext.js
+++ b/src/ext.js
@@ -206,7 +206,7 @@ export default function ext({ translator, shouldHide, senseNavigation, theme, is
                   expression: "optional",
                   ref: "navigation.chart",
                   translation: "properties.chartPicker.chart",
-                  show: (data) => isChartNavigation ? checkShowNavigation(data, "sheet", "chart") : false,
+                  show: (data) => (isChartNavigation ? checkShowNavigation(data, "chart") : false),
                 },
                 story: {
                   type: "string",
--- a/src/ext.js
+++ b/src/ext.js
@@ -21,7 +21,7 @@ const getAutomations = async () => {
   }
   return automationsList;
 };
-export default function ext({ translator, shouldHide, senseNavigation, theme }) {
+export default function ext({ translator, shouldHide, senseNavigation, theme, isChartNavigation }) {
   const multiUserAutomation =
     shouldHide.isEnabled && shouldHide.isEnabled("SENSECLIENT_IM_1855_AUTOMATIONS_MULTI_USER");
   const stylingPanelEnabled = shouldHide.isEnabled && shouldHide.isEnabled("SENSECLIENT_IM_1525_STYLINGPANEL_BUTTON");
@@ -206,7 +206,7 @@ export default function ext({ translator, shouldHide, senseNavigation, theme }) 
                   expression: "optional",
                   ref: "navigation.chart",
                   translation: "properties.chartPicker.chart",
-                  show: (data) => checkShowNavigation(data, "sheet", "chart"),
+                  show: (data) => isChartNavigation ? checkShowNavigation(data, "sheet", "chart") : false,
                 },
                 story: {
                   type: "string",
--- a/src/ext.js
+++ b/src/ext.js
@@ -21,7 +21,7 @@ const getAutomations = async () => {
   }
   return automationsList;
 };
-export default function ext({ translator, shouldHide, senseNavigation, theme, isChartNavigation }) {
+export default function ext({ translator, shouldHide, senseNavigation, theme, isGoToChartEnabled }) {
   const multiUserAutomation =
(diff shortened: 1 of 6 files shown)


cbt1:
You appear to be adding a new feature. Then this is a `feat` PR and not `chore`.

Donya-qlik:
> You appear to be adding a new feature. Then this is a `feat` PR and not `chore`.

Ok, I'll change it. Do you think I need a FF ? I've a related change in sense-client at the same time.

haxxmaxx:
So have we decided that the way you can go to an object is by the extra option under go to sheet? Not a separate action called go to object?

Donya-qlik:
> So have we decided that the way you can go to an object is by the extra option under go to sheet? Not a separate action called go to object?

Yes, it has been decided to be an extra option for now.

src/object-properties.js
```diff
@@ -132,7 +132,8 @@ const properties = {
  * @typedef {object} NavigationAction
  * @property {'nextSheet'|'prevSheet'|'firstSheet'|'lastSheet'|'goToSheet'|'goToSheetById'|'goToSheetById'|'goToStory'|'openWebsite'|'odagLink'|'openChainedApp'} action - Name of navigation action
  * @property {string=} appId - app ID. Required for 'openChainedApp'
- * @property {string=} sheet - sheet ID. Required for 'goToSheet' and 'goToSheetById'
+ * @property {string=} sheet - sheet - sheet ID. Required for 'goToSheet' and 'goToSheetById'
+ * @property {string=} chart - chart - chart ID. Required for 'goToSheet' and specify the chart
```
haxxmaxx, line 136: I don't think it is required for goToSheet right? you can still just go to a sheet and the if you want to you specify a chart as well? Also looks like a typo, it says chart - chart, and sheet - sheet

src/object-properties.js
```diff
@@ -132,7 +132,8 @@ const properties = {
  * @typedef {object} NavigationAction
  * @property {'nextSheet'|'prevSheet'|'firstSheet'|'lastSheet'|'goToSheet'|'goToSheetById'|'goToSheetById'|'goToStory'|'openWebsite'|'odagLink'|'openChainedApp'} action - Name of navigation action
  * @property {string=} appId - app ID. Required for 'openChainedApp'
- * @property {string=} sheet - sheet ID. Required for 'goToSheet' and 'goToSheetById'
+ * @property {string=} sheet - sheet - sheet ID. Required for 'goToSheet' and 'goToSheetById'
```
haxxmaxx, line 135: Here it says that there is a goToSheetById as well, we are not adding the go to chart option there?

src/utils/navigation-actions.js
```diff
@@ -52,15 +52,13 @@ const navigationActions = [
...
-
   {
     translation: "Object.ActionButton.GoToASheet",
     value: "goToSheet",
-    navigationCall: async ({ senseNavigation, sheet }) => {
-      sheet && (await senseNavigation.goToSheet(sheet));
+    navigationCall: async ({ senseNavigation, sheet, chart = "" }) => {
+      sheet && (await senseNavigation.goToSheet(sheet, chart));
     },
-    // TODO replace by searchable dropdown
```
haxxmaxx, line 62: why remove this comment?

src/ext.js
```diff
@@ -201,6 +201,13 @@ export default function ext({ translator, shouldHide, senseNavigation, theme })
                     }));
                   },
                 },
+                chart: {
+                  type: "string",
+                  expression: "optional",
+                  ref: "navigation.chart",
+                  translation: "properties.chartPicker.chart",
+                  show: (data) => checkShowNavigation(data, "sheet", "chart"),
```
haxxmaxx, line 209: How come are you adding a third argument here? `checkShowNavigation` only has two arguments

src/object-properties.js
Donya-qlik: I saw the other items from navigations are all added so added this one as well. 
Sorry for my messy PR, should clean that up before put it for review. All comments are fixed.

haxxmaxx: I mean you will need to add it here, but the chart prop should not say that it is required. For now though I think it is good that we don't change the API spec, you can do that in a different PR

src/index.js
```diff
@@ -26,6 +26,7 @@ export default function supernova(env) {
   const isUnsupportedFeature = anything.sense?.isUnsupportedFeature;
   const shouldHide = { isEnabled, isFeatureBlacklisted, isUnsupportedFeature };
   const multiUserAutomation = isEnabled?.("SENSECLIENT_IM_1855_AUTOMATIONS_MULTI_USER");
+  const isChartNavigation = isEnabled?.("ACTION_BUTTON_IM_4975_CHART_NAVIGATION");
```
haxxmaxx, line 29: I think this could have a better name. this sounds like you can navigate inside a chart. `isGoToChartEnabled` or something similar is more descriptive IMO

haxxmaxx (approved):

src/index.js
Donya-qlik: changed it to `isGoToChartEnabled`

haxxmaxx (approved):

src/utils/navigation-actions.js
```diff
@@ -52,15 +52,14 @@ const navigationActions = [
...
     requiredInput: [],
   },
-
   {
     translation: "Object.ActionButton.GoToASheet",
     value: "goToSheet",
-    navigationCall: async ({ senseNavigation, sheet }) => {
-      sheet && (await senseNavigation.goToSheet(sheet));
+    navigationCall: async ({ senseNavigation, sheet, chart = "" }) => {
+      sheet && (await senseNavigation.goToSheet(sheet, chart));
```
niekvanstaveren, line 59: Why defaults this one to a string? Should be able to be undefined I think.

PR #443, opened 2023-11-07
fix: undefined background property

Donya-qlik:
It fixes the console type error of:

image

### DIFF
--- a/src/hooks/use-load-image.js
+++ b/src/hooks/use-load-image.js
@@ -11,7 +11,7 @@ const imagePromise = (url) =>
   });
 
 const useLoadImage = (layout) => {
-  const { url } = layout.style.background;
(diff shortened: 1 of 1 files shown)


src/hooks/use-load-image.js
```diff
@@ -11,7 +11,7 @@ const imagePromise = (url) =>
   });
 
 const useLoadImage = (layout) => {
-  const { url } = layout.style.background;
+  const url = layout.style?.background?.url ?? undefined;
```
cbt1, line 14: Don't need to do `?? undefined`, but maybe you are getting a lint error?

Donya-qlik: True, no I don't get lint error, removed it.

cbt1 (approved):

PR #445, opened 2023-11-07
feat: rename navigationCall parameter for chartId

Donya-qlik:


src/object-properties.js
```diff
@@ -133,6 +133,7 @@ const properties = {
  * @property {'nextSheet'|'prevSheet'|'firstSheet'|'lastSheet'|'goToSheet'|'goToSheetById'|'goToSheetById'|'goToStory'|'openWebsite'|'odagLink'|'openChainedApp'} action - Name of navigation action
  * @property {string=} appId - app ID. Required for 'openChainedApp'
  * @property {string=} sheet - sheet ID. Required for 'goToSheet' and 'goToSheetById'
+ * @property {string=} chartId - chart ID. Required for highlighting the chart in the sheet
```
haxxmaxx, line 136: You need to specify when this is used, namely for the goToSheet action

api-specifications/properties.json
```diff
@@ -351,6 +351,11 @@
             "optional": true,
             "type": "string"
           },
+          "chartId": {
```
haxxmaxx, line 354: This is bit strange, but I'm pretty sure we actually update this file when we release, not when we make changes to object-properties. @niekvanstaveren do you remember if this is true? In that case, this should be reverted

Donya-qlik: This is an API specification error I got in CI so here is the result of running `yarn spec`

niekvanstaveren: No the workflow will fail when the api specification is not updated properly. The version of this needs to be updated when released, but the specifications should already be updated. 
This PR should however not be a refactor. You cannot do a refactor and change the API

Donya-qlik: @niekvanstaveren ok didn't know that, should I change it to `chore`?

niekvanstaveren: API changes should probably always be `feat`. In very rare occasions I guess it could be a `fix`, but that would only be when you are fixing something in the API

src/ext.js
```diff
@@ -201,12 +201,12 @@ export default function ext({ translator, shouldHide, senseNavigation, theme, is
...
                   },
                 },
-                chart: {
+                chartId: {
                   type: "string",
                   expression: "optional",
-                  ref: "navigation.chart",
-                  translation: "properties.chartPicker.chart",
-                  show: (data) => (isGoToChartEnabled ? checkShowNavigation(data, "chart") : false),
+                  ref: "navigation.chartId",
```
cbt1, line 207: Are there any sn-action-button that already exist with the `navigation.chart` reference? If so, is there not a risk this will break those charts?

Donya-qlik: No, not any more reference, this s a new feature, the other part of it that makes it work in in sense-client [here](https://github.com/qlik-trial/sense-client/pull/25579)

niekvanstaveren (approved):

PR #451, opened 2023-12-04
feat: set a dropdown of charID options for navigation to a chart

Donya-qlik:


src/ext.js
```diff
@@ -203,10 +203,18 @@ export default function ext({ translator, shouldHide, senseNavigation, theme, is
...
-                  expression: "optional",
                   ref: "navigation.chartId",
                   translation: "properties.chartPicker.chartId",
+                  component: "expression-with-dropdown",
+                  expressionType: "StringExpression",
                   show: (data) => (isGoToChartEnabled ? checkShowNavigation(data, "chartId") : false),
+                  options: async (data, action, hyperCubeHandler) => {
+                    const sheetId = data.navigation.sheet;
+                    const objects = sheetId && (await hyperCubeHandler.app.getObject(sheetId));
+                    return objects?.properties.cells.map((cell) => ({
```
haxxmaxx, line 214: For clarity, this should be called just `object` or even `sheet`, to no be confused with the objects _on_ the sheet.

src/__tests__/ext.spec.js
```diff
@@ -113,6 +113,21 @@ describe("ext", () => {
         },
       },
     ];
+    const chartIds = {
```
haxxmaxx, line 116: same here, this isn't a list of chart id, it is a sheet object, that has a list of cells

haxxmaxx (approved):

src/ext.js
Donya-qlik: Umm... here I'm getting the objects on the sheet. If the sheet has three charts that means it contains three objects, doesn't it ?

Donya-qlik: Ok, I'm thinking, you're right it's actually a sheet. I'll name it 'sheet'.

PR #454, opened 2023-12-12
feat: step cyclic dimension with sn-action-button

Bjornwa:
Add cyclic group action under feature flag.
Note: backend called is not working until engine release the new API.


src/utils/__tests__/actions.spec.js
```diff
@@ -576,7 +576,7 @@ describe("actions", () => {
         isEnabled: jest.fn().mockReturnValue(true),
       };
       const result = getActionsList(shouldHide);
-      expect(result.length).toBe(20);
+      expect(result.length).toBe(21);
```
linhnguyen-qlik, line 579: What does unsupported feature navigations means? Is cyclic group an unsupported feature navigation?

src/utils/__tests__/actions.spec.js
```diff
@@ -585,7 +585,7 @@ describe("actions", () => {
         isEnabled: jest.fn().mockReturnValue(true),
       };
       const result = getActionsList(shouldHide);
-      expect(result.length).toBe(13);
+      expect(result.length).toBe(14);
```
linhnguyen-qlik, line 588: What does `feature blacklisted navigations` means? is cyclic group one of the feature blacklisted navigations?

src/utils/__tests__/actions.spec.js
niekvanstaveren: This is for apps running in Direct Query mode. Unsupported features are controlled by engine and stored on the app layout. It would be good to verify if cyclic groups is something that can or cannot work in DQ apps

src/utils/__tests__/actions.spec.js
niekvanstaveren: This is also for apps running in Direct Query mode. The difference with unsupported features is that this is controlled by the client. Mostly bigger parts that just do not make sense when the app is DQ

src/utils/__tests__/actions.spec.js
linhnguyen-qlik: Ok. Björn have noted this in our test plan. I guess in the mean time we let this test be like this until we know it has support or not

linhnguyen-qlik:
> Maybe some tests would be nice

Tests added :)

linhnguyen-qlik (approved):

PR #455, opened 2023-12-12
feat: evaluate expression at runtime

niekvanstaveren:
Adding an options to have expression evaluate at runtime:

image

When this option is set the actions will run in order, but instead of using the expression values from the layout the expressions will be evaluated just before the action is run. This means that previous actions can impact the state for the next actions.


src/components/action-button.js
```diff
@@ -1,11 +1,24 @@
...
-export const runActions = async (actionList) => {
-  for (let i = 0; i < actionList.length; i++) {
-    // eslint-disable-next-line no-await-in-loop
-    await actionList[i]();
+export const runActions = async ({ actionCallList, model, layout, app }) => {
+  if (layout.runtimeExpressionEvaluation) {
+    const properties = await model.getProperties();
+    for (let i = 0; i < actionCallList.length; i++) {
+      const overrideValue =
+        properties.actions[i].value?.qStringExpression?.qExpr &&
```
cbt1, line 11: I have no idea what `actionCallList` does but, do note that `overrideValue` can become any falsy value. For example if `qExpr === ''`, `overrideValue` becomes an empty string.

Same for `overrideVariable`.

cbt1: Is `actions` a required property on `properties`?

cbt1: And does it always match the length of `actionCallList`

src/components/action-button.js
```diff
@@ -1,11 +1,24 @@
...
+        (await app.evaluate(properties.actions[i].value?.qStringExpression?.qExpr));
+      const overrideVariable =
+        properties.actions[i].variable?.qStringExpression?.qExpr &&
+        (await app.evaluate(properties.actions[i].variable?.qStringExpression?.qExpr));
```
haxxmaxx, line 15: you can remove optional chaining here, a bit clearer I would say

src/ext.js
```diff
@@ -33,164 +33,175 @@ export default function ext({ translator, shouldHide, senseNavigation, theme, is
...
+                      defaultValue: false,
+                      show: (data) => checkShowAction(data, "partial"),
+                    },
+                    automationProps: {
+                      type: "items",
+                      grouped: false,
+                      items: getAutomationProps(multiUserAutomation, getAutomations),
+                      show: (data) => checkShowAction(data, "automation"),
+                    },
+                    cyclicGroup: {
```
haxxmaxx, line 167: this is just moved right?

src/components/__tests__/action-button.spec.js
```diff
@@ -159,12 +159,12 @@ describe("action button", () => {
   });
   describe("runActions", () => {
```
haxxmaxx, line 160: Shouldn't there be a new test, so we have one with the prop off and one with it on?

src/utils/actions.js
```diff
@@ -140,11 +141,12 @@ const actions = [
     group: "selection",
     getActionCall:
       ({ app, qStateName, field, value }) =>
-      async () => {
-        if (field && value) {
+      async (overrideValue) => {
+        const useValue = overrideValue || value;
```
haxxmaxx, line 145: yeah as Christofer mentioned, this could fail if overrideValue is `0` or `""`
```suggestion
        const useValue = overrideValue ?? value;
```

api-specifications/properties.json
```diff
@@ -70,6 +70,11 @@
             "type": "#/definitions/Action"
           }
         },
+        "runtimeExpressionEvaluation": {
+          "description": "Boolean to determine the parallel evaluation of expression. Set to true if you want the expression to be evaluated between the actions (instead of upfront).",
```
haxxmaxx, line 74: So with this off it all expressions are evaluated when you press the button, right? Then I would word this something like this
```suggestion
          "description": "Boolean to determine the timing for evaluating expressions. Set to true if you want the expression to be evaluated before each action (instead of evaluating all when button is pressed).",
```

haxxmaxx (commented):
Probably change the title and a short description

src/components/action-button.js
niekvanstaveren: That is a couple of questions. In order:

1. ActionCallList is the list of actions that need to run. Actions from actions get pushed into this array if they are found. Good catch. I'll look at that
2. Yes it is. 
3. It does not have to match actually. Since we first try to find the actions from a list the actionCallList might be shorter than actions. This could happen if you create the button through the api. It also makes sure it is backwards compatible. Most important though is probably conversion from the old button. There are some actions that we do not support in the new if I remember correct

src/components/__tests__/action-button.spec.js
niekvanstaveren: Yes

src/ext.js
niekvanstaveren: Yes. Needed to add the toggle on the same level as the actions. So this has all moved down a bit:

image

src/components/action-button.js
niekvanstaveren: Fixed based on suggestion from Max. Also made sure that the actionCallList and actions list have the same length

cbt1 (approved):
No other comments from my side

api-specifications/properties.json
haxxmaxx: I mean at least change this, upfront means something else
```suggestion
          "description": "Boolean to determine the parallel evaluation of expression. Set to true if you want the expression to be evaluated between the actions (instead of up front).",
```

haxxmaxx (approved):

PR #457, opened 2023-12-13
chore: use pnpm as package manager

niekvanstaveren:


cbt1 (approved):
LGTM, assuming that you built and tested it a bit locally as well

niekvanstaveren:
> LGTM, assuming that you built and tested it a bit locally as well

I did yes. Will do another round when actually bumping in client.

PR #478, opened 2024-03-06
chore: disable renovate from bumping peerDependencies

haxxmaxx:


renovate.json
```diff
@@ -7,7 +7,7 @@
   "rangeStrategy": "bump",
   "packageRules": [
     {
-      "depTypeList": ["engines"],
+      "matchDepTypes": ["engines", "peerDependencies"],
```
haxxmaxx, line 10: new name for depTypeList

package.json
```diff
@@ -109,7 +109,7 @@
     "yargs": "17.7.2"
   },
   "peerDependencies": {
-    "@nebula.js/stardust": ">=4.11.0"
+    "@nebula.js/stardust": ">=4.x"
```
cbt1, line 112: I might need to think more about this but, we cannot know if version >=5 of stardust is supported until it has been released.

What would happen if v5 is released and we don't support it? I would think that means any previous release now have a "faulty" version in peer deps. Fixing it would require us to make a new release with an updated peer deps version, which itself could be a breaking change if both v4 and v5 ends up not being supported.

The same logic could apply to minor versions as well. So we might say `^4.x` is supported but then we need `4.15.0` and `^4.x` would no longer be valid.

So safest is likely to set it as `^4.11.0`, bump it only when a new feature is needed and set is a breaking change.

haxxmaxx: that makes sense. What does `^` mean in the case of a peer dependency? Wouldn't it be the same as `>=`, as in pick the latest one, but the one listed is the earliest one working?

cbt1: I think it means pick the latest package version that satisfies `>=4.x`, which would mean any package greater or equal to `4.0.0`. So if `12.37.0` exist, we are telling the consumers of `sn-action-button` that version `12.37.0` of stardust is an OK version to use.

I've been using https://semver.npmjs.com/#syntax-examples to test my version range.

haxxmaxx: ok so >=4.x means any version from 4.0.0 and above, while ^4.11.0 would mean any minor version equal or higher than 4.11, but not any major versions? then that makes sense here

cbt1 (approved):

